PR #11, opened 2021-04-18
feat: support linux installation flow

dimabru:


install.sh
```diff
@@ -10,7 +12,13 @@ OUTPUT_BASENAME_WITH_POSTFIX=$OUTPUT_BASENAME.zip
 curl -L $DOWNLOAD_URL -o $OUTPUT_BASENAME_WITH_POSTFIX
 unzip $OUTPUT_BASENAME_WITH_POSTFIX -d $OUTPUT_BASENAME
 
-cp $OUTPUT_BASENAME/datree /usr/local/bin
+if [[ $osName == "Linux" ]]
+then
+    sudo cp $OUTPUT_BASENAME/datree /usr/local/bin
+    mkdir ~/.datree
```
myishay, line 18: Is the `mkdir` required only in Linux? Not in Darwin?
Either way, couldn't be bad making this dir in advance on Darwin too

myishay (approved):

install.sh
dimabru: no need to do this in Darwin. I prefer to keep this behavior only when it is required to avoid adding more dependencies where we don't need to

PR #15, opened 2021-04-19
feat: support yaml and json output

dimabru:
* Add flag `-o, --output` to `datree test` to allow output in different formats 
* currently suppot `json` and `yaml`

### DIFF
--- a/bl/evaluator_test.go
+++ b/bl/evaluator_test.go
@@ -252,7 +252,7 @@ func TestPrintResults(t *testing.T) {
 		}{},
 	}
 
-	evaluator.PrintResults(&results, "cli_id")
+	evaluator.PrintResults(&results, "cli_id", "")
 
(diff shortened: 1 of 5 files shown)


bl/evaluator_test.go
```diff
@@ -252,7 +252,7 @@ func TestPrintResults(t *testing.T) {
 		}{},
 	}
 
-	evaluator.PrintResults(&results, "cli_id")
+	evaluator.PrintResults(&results, "cli_id", "")
```
alexfedin, line 255: I think we should add tests to check that the other formats output works as expected

alexfedin (changes requested):

bl/evaluator_test.go
dimabru: added tests

PR #16, opened 2021-04-19
fix: send log metadata

myishay:


bl/osInfo.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "github.com/shirou/gopsutil/host"
+
+type OSInfo struct {
+	OS              string
+	PlatformVersion string
+	KernelVersion   string
+}
+
+func createOSInfo() *OSInfo {
```
alexfedin: the convention for constructors is NewOsInfo

bl/evaluator.go
```diff
@@ -55,7 +64,30 @@ func (e *Evaluator) Evaluate(pattern string, cliId string, evaluationConc int) (
...
+
+	for _, file := range files {
+		filesProperties = append(filesProperties, *file)
+	}
+
+	evaluationRequest := cliClient.EvaluationRequest{
+		CliId:   cliId,
+		Pattern: pattern,
+		Metadata: struct {
+			CliVersion      string "json:\"cliVersion\""
```
alexfedin: I don't think that the json mapping should be used outside of cliClient , this is part of the contract between the client and the server and it should be irrelevant for the rest of the application.

if another place will use that function in the future they will have to do the same parsing

You can use another object to wrap everything you need for the request and the cli service will create the request object from the one you added
or 
You can add a factory function to cliClient to create the evaluation request for you and you will pass it all the values it needs

alexfedin (changes requested):

bl/evaluator.go
myishay: Which place could use this struct?
Either way, this struct describes a part of the signature of `RequestEvaluation`, and every place that would use this function has to use this type

alexfedin (approved):

PR #22, opened 2021-04-28
fix: total failed count

dimabru:
* Fixed issue where total failed rules count is displaying unique rules instead of total aggregated amount of fails per file

### DIFF
--- a/bl/evaluator.go
+++ b/bl/evaluator.go
@@ -148,12 +148,8 @@ func (e *Evaluator) aggregateEvaluationResults(evaluationResults []cliClient.Eva
 
 	totalRulesCount := len(evaluationResults)
 	totalFailedCount := 0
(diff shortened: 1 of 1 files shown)


bl/evaluator.go
```diff
@@ -162,17 +158,12 @@ func (e *Evaluator) aggregateEvaluationResults(evaluationResults []cliClient.Eva
 
 			// file and rule not already exists in mapper
 			if _, exists := mapper[match.FileName][result.Rule.ID]; !exists {
-				filenames[match.FileName] = match.FileName
+				totalFailedCount++
```
myishay, line 161: in case we have more than 1 occurrences in a file, this will cause more fails than expected

dimabru: no it won't :)

PR #26, opened 2021-05-05
fix: remove patten logic

dimabru:
* Removed dependencies in external glob packages
* Path now supports simple `datree test <path>` without the need for parentheses
* Removed `pattern` from the request content to the cli service

### DIFF
--- a/cmd/test/main.go
+++ b/cmd/test/main.go
@@ -18,7 +18,7 @@ type LocalConfigManager interface {
 type Evaluator interface {
 	PrintResults(results *bl.EvaluationResults, cliId string, output string) error
 	PrintFileParsingErrors(errors []propertiesExtractor.FileError)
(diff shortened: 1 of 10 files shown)


cmd/test/main.go
```diff
@@ -19,7 +18,7 @@ type LocalConfigManager interface {
 type Evaluator interface {
 	PrintResults(results *bl.EvaluationResults, cliId string, output string) error
 	PrintFileParsingErrors(errors []propertiesExtractor.FileError)
-	Evaluate(pattern string, cliId string, evaluationConc int, cliVersion string) (*bl.EvaluationResults, []propertiesExtractor.FileError, error)
+	Evaluate(patterns []string, cliId string, evaluationConc int, cliVersion string) (*bl.EvaluationResults, []propertiesExtractor.FileError, error)
```
myishay: rename to `paths`

dimabru: done

PR #30, opened 2021-05-09
feat: print version message

myishay:


pkg/printer/printer.go
```diff
@@ -94,6 +94,24 @@ func (p *Printer) PrintSummaryTable(summary Summary) {
 }
 
 func (p *Printer) printInColor(title string, color *color.Color) {
-	warningColorPrintFn := color.PrintfFunc()
-	warningColorPrintFn(title)
+	colorPrintFn := color.PrintfFunc()
+	colorPrintFn(title)
+}
+
+func (p *Printer) createNewColor(clr string) *color.Color {
```
noaabarki, line 101: nice

noaabarki (changes requested):

pkg/cliClient/cliClient.go
```diff
@@ -77,3 +83,18 @@ func (c *CliClient) RequestEvaluation(request EvaluationRequest) (EvaluationResp
 
 	return *evaluationResponse, nil
 }
+
+func (c *CliClient) GetVersionMessage(ch chan VersionMessage, cliVersion string) {
+	defer close(ch)
```
noaabarki: I would keep `GetVersionMessage` primitive => call endpoint and returns an answer
the channel is more of a BL plus rule of thumb to remember - the one that creates a channel should also close it

main.go
```diff
@@ -1,7 +1,26 @@
...
 func main() {
-	cmd.Execute()
+	c := cliClient.NewCliClient(deploymentConfig.URL)
+	p := printer.CreateNewPrinter()
+
+	messageChannel := make(chan cliClient.VersionMessage, 1)
+	go c.GetVersionMessage(messageChannel, cmd.CliVersion)
+	err := cmd.Execute()
+	msg := <-messageChannel
+	if msg.MessageText != "" {
```
noaabarki: Let's use select

pkg/printer/printer.go
```diff
@@ -30,15 +30,15 @@ type Warning struct {
 
 func (p *Printer) PrintWarnings(warnings []Warning) {
 	for _, warning := range warnings {
-		p.printInColor(warning.Title, p.theme.Colors.Warning)
+		p.printInColor(warning.Title, p.theme.Colors.Yello)
```
dimabru: Yellow

pkg/printer/printer.go
```diff
@@ -94,6 +94,24 @@ func (p *Printer) PrintSummaryTable(summary Summary) {
...
-	warningColorPrintFn(title)
+	colorPrintFn := color.PrintfFunc()
+	colorPrintFn(title)
+}
+
+func (p *Printer) createNewColor(clr string) *color.Color {
+	switch clr {
+	case "red":
+		return p.theme.Colors.Red
+	case "yello":
```
dimabru: yellow

pkg/printer/printer.go
```diff
@@ -94,6 +94,24 @@ func (p *Printer) PrintSummaryTable(summary Summary) {
...
+		return p.theme.Colors.Yello
+	case "green":
+		return p.theme.Colors.Green
+	default:
```
dimabru, line 109: we should also have case for explicit white color. in case we decide to change the default later on (backend supports white as well)

myishay: what should be the default then?

dimabru: default can also be white. but there's a difference between getting explicit white and adding white as fallback.
so both case and default can be white, but should we decide to change the default, explicit white from backend will still be supported

bl/versionMessage.go
```diff
@@ -0,0 +1,34 @@
...
+	"time"
+
+	"github.com/datreeio/datree/pkg/cliClient"
+	"github.com/datreeio/datree/pkg/deploymentConfig"
+	"github.com/datreeio/datree/pkg/printer"
+)
+
+func PopulateVersionMessageChan(cliVersion string) chan *cliClient.VersionMessage {
+	messageChannel := make(chan *cliClient.VersionMessage, 1)
+	go func() {
```
noaabarki, line 13:
```
func PopulateVersionMessageChan(...) chan *cliClient.VersionMessage {
  messageChannel := make(chan *string, 1)
	  go func() {
                   ....
		  close(messageChannel)
	  }()
...
}
```

```
func HandleVersionMessage(...) {
        msg, ok := <-messageChannel
	if(ok) {
		fmt.Println("msg:", msg)
	}
}
```

noaabarki (approved):

PR #34, opened 2021-05-16
Dat 3230 validate k8s (wip)

noaabarki:


bl/evaluation/printer.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/datreeio/datree/pkg/printer"
+	"gopkg.in/yaml.v2"
+)
+
+type Printer interface {
+	PrintWarnings(warnings []printer.Warning)
+	PrintSummaryTable(summary printer.Summary)
+}
+
+// url := "https://app.datree.io/login?cliId=" + cliId
```
dimabru, line 24: comment

bl/evaluation/printer.go
```diff
@@ -0,0 +1,138 @@
...
+func PrintResults(results *EvaluationResults, loginURL string, outputFormat string, printer Printer) error {
+	switch {
```
dimabru, line 26: love this!!

bl/validator/validator.go
```diff
@@ -0,0 +1,96 @@
+package validator
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+
+	"github.com/datreeio/datree/bl/files"
+	kubeconformValidator "github.com/yannh/kubeconform/pkg/validator"
```
dimabru: is there a way to specify specific commit here? to try and keep it more stable

cmd/version/main.go
```diff
@@ -3,22 +3,34 @@ package version
...
-	CliVersion           string
-	VersionMessageClient bl.VersionMessageClient
+	CliVersion string
+	Messager   Messager
+	Printer    Printer
 }
 
 func version(ctx *VersionCommandContext) {
-	messageChannel := bl.PopulateVersionMessageChan(ctx.VersionMessageClient, ctx.CliVersion)
+	messages := make(chan *messager.VersionMessage, 1)
```
dimabru, line 24: we can only have one message. I think this assumes that each version supports multiple messages which we don't at the moment.
Not a big issue, just pointing it out

pkg/cliClient/evaluation.go
```diff
@@ -0,0 +1,100 @@
...
+	KernelVersion   string `json:"kernelVersion"`
+}
+
+type CreateEvaluationRequest struct {
+	CliId    string    `json:"cliId"`
+	Metadata *Metadata `json:"metadata"`
+}
+
+func (c *CliClient) CreateEvaluation(request *CreateEvaluationRequest) (int, error) {
+	httpRes, err := c.httpClient.Request(http.MethodPost, "/cli/evaluate/create", request, nil)
```
dimabru: request url is `cli/evaluation/create`
Yishay's fix

pkg/cliClient/evaluation.go
```diff
@@ -0,0 +1,100 @@
...
+		return &EvaluationResponse{}, err
+	}
+
+	return evaluationResponse, nil
+}
+
+type UpdateEvaluationValidationRequest struct {
+	EvaluationId   int       `json:"evaluationId"`
+	InvalidFiles   []*string `json:"failedFiles"`
+	StopEvaluation bool
```
dimabru: missing json expected value `stopEvaluation`

cmd/test/main.go
```diff
@@ -56,38 +67,56 @@ func NewTestCommand(ctx *TestCommandContext) *cobra.Command {
...
-	s.Color("cyan")
+	s := createSpinner(" Loading...", "cyan")
 	s.Start()
 
-	messageChannel := bl.PopulateVersionMessageChan(ctx.VersionMessageClient, ctx.CliVersion)
+	messages := make(chan *messager.VersionMessage, 1)
+	go ctx.Messager.LoadVersionMessages(messages, ctx.CliVersion)
 
-	config, err := ctx.LocalConfig.GetConfiguration()
+	validPaths, invalidPaths, _ := ctx.Validator.Validate(paths)
```
dimabru: kubeconform step is performed before the `createEvaluation` request. although it doesn't make much difference right now because they are unrelated steps, I feel like conseptually we should run the validation step after we create an evaluation.
In case there's a panic inside validate or whatever, this might break the flow without any record of it being run.
also, we will later add another step (yamlValidation) and I think we should keep the step 0 -> 1 -> 2 -> 3 structure

cmd/test/main.go
```diff
@@ -56,38 +67,56 @@ func NewTestCommand(ctx *TestCommandContext) *cobra.Command {
...
+
+	evaluationId, err := ctx.Evaluator.CreateEvaluation(ctx.LocalConfig.CliId, ctx.CliVersion)
 	if err != nil {
 		fmt.Println(err.Error())
 		return err
 	}
 
-	evaluationResponse, fileParsingErrors, err := ctx.Evaluator.Evaluate(paths, config.CliId, 50, ctx.CliVersion)
+	results, errors, err := ctx.Evaluator.Evaluate(validPaths, invalidPaths, evaluationId)
```
dimabru: you chose here to forward the `invalidPaths` array into the evaluate function and let the evaluator deal with it. 
note that if we're giving the evaluator the 'power' to handle invalid files, we will need to later enhance it once we add the second validation - yaml validation (step 1).
not a deal breaker, just pointing that out.

dimabru (changes requested):

cmd/version/main.go
myishay: as much as I know this is the best way to handle 1 buffered channels (the last argument `1` indicates that this channel is allocated to only one member)

PR #46, opened 2021-05-24
fix: handle unknown files + load version message in advance

myishay:


cmd/test/main.go
```diff
@@ -87,14 +87,19 @@ func New(ctx *TestCommandContext) *cobra.Command {
...
 		msg, ok := <-messages
 		if ok {
 			ctx.Printer.PrintMessage(msg.MessageText+"\n", msg.MessageColor)
 		}
 	}()
 
-	filePaths := ctx.Reader.FilterFiles(paths)
+	filePaths, err := ctx.Reader.FilterFiles(paths)
+	if err != nil {
+		fmt.Println(err.Error())
```
alexfedin, line 100: we should add a more verbose error here, like what failed

cmd/test/main_test.go
```diff
@@ -80,9 +80,9 @@ type ReaderMock struct {
 	mock.Mock
 }
 
-func (rm *ReaderMock) FilterFiles(paths []string) []string {
+func (rm *ReaderMock) FilterFiles(paths []string) ([]string, error) {
```
alexfedin, line 83: missing test where filter files returns an error

alexfedin (changes requested):

cmd/test/main.go
myishay: actually in this case the error is very straight forward: `no such file or directory`

PR #54, opened 2021-05-31
feat: display configs count

dimabru:


cmd/test/main.go
```diff
@@ -166,6 +168,16 @@ func test(ctx *TestCommandContext, paths []string, flags TestCommandFlags) error
 	return invocationFailedErr
 }
 
+func countConfigurations(filesConfigurations []*extractor.FileConfiguration) int {
```
alexfedin, line 171: I think we should move this into the format function in evaluator if no one else is using filesConfigurations parameter down the line

alexfedin (changes requested):

bl/evaluation/printer.go
```diff
@@ -145,7 +145,7 @@ func (t OutputTitle) String() string {
 		"Total rules failed"}[t]
 }
 func parseEvaluationResultsToSummary(results *EvaluationResults, evaluationSummary printer.EvaluationSummary, loginURL string) printer.Summary {
-	filesCount := evaluationSummary.FilesCount
+	configsCount := evaluationSummary.ConfigsCount
 	rulesCount := 0
 	totalRulesEvaluated := 0
```
alexfedin, line 150: test comment

alexfedin (approved):

cmd/test/main.go
dimabru: `filesConfigurations` is used below

PR #55, opened 2021-06-01
Dat 3235 send yaml validation

myishay:


dimabru (dismissed):

bl/evaluation/evaluator_test.go
```diff
@@ -147,14 +138,13 @@ type evaluateTestCase struct {
 }
 
 func request_evaluation_all_valid() *evaluateTestCase {
-	invalidPath := "path/path1/service.yaml"
+	validFilePath := "../../internal/fixtures/kube/pass-all.yaml"
```
noaabarki, line 144: I think we should add more tests

pkg/printer/printer.go
```diff
@@ -27,6 +28,7 @@ type WarningInfo struct {
 
 type ValidationInfo struct {
 	IsValid          bool
+	ValidationStatus validation.ValidationStatus
```
noaabarki, line 31: I suggest rename to 'isValid, status, errors, schemaVersion'

noaabarki (changes requested):

PR #60, opened 2021-06-06
docs: add alpha release warning

romanlab:


eyarz (changes requested):
add something about the fact that we are doing our best to be backward compatible and we have deprecation notice in the CLI

eyarz (approved):

PR #63, opened 2021-06-09
fix: rules count in summary

dimabru:


bl/evaluation/evaluator.go
```diff
@@ -76,10 +76,19 @@ func (e *Evaluator) UpdateFailedK8sValidation(invalidFiles []*validation.Invalid
 	return err
 }
 
-func (e *Evaluator) Evaluate(filesConfigurations []*extractor.FileConfigurations, evaluationId int) (*EvaluationResults, error) {
+func (e *Evaluator) Evaluate(filesConfigurations []*extractor.FileConfigurations, evaluationId int, rulesCount int) (*EvaluationResults, error) {
```
alexfedin: it looks like you only pass rulesCount in order to put it in the response object you get from this function.
in that case I don't think rulesCount belongs here, as this function has no use for it to do the evaluation.
I suggest removing the rulesCount from the evaluationResults object entirely as it belongs to the evaluation creation rather than evaluation response.

alexfedin (changes requested):

myishay (dismissed):

bl/evaluation/evaluator.go
dimabru: moved `RulesCount` to evaluation summary object

PR #65, opened 2021-06-14
feat: config set token

dimabru:
* Added config cmd package to deal with config commands
* Refactored `localConfig`
* Support command `datree config set [key] [value]` - changes key-value in config.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,3 +12,6 @@ create-bin:
 
 print-version:
 	go run -tags=staging -ldflags="-X github.com/datreeio/datree/cmd.CliVersion=0.0.1" main.go version
+
+set-token:
+	go run -tags=staging -ldflags="-X github.com/datreeio/datree/cmd.CliVersion=0.0.1" main.go config set token testtoken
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -12,3 +12,6 @@ create-bin:
 
 print-version:
 	go run -tags=staging -ldflags="-X github.com/datreeio/datree/cmd.CliVersion=0.0.1" main.go version
+
+set-token:
```
myishay, line 16: Alt Text

cmd/config/set.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+
+	"github.com/datreeio/datree/bl/messager"
+	"github.com/spf13/cobra"
+)
+
+func NewSetCommand(ctx *ConfigCommandContext) *cobra.Command {
+	setCommand := &cobra.Command{
```
myishay, line 11: I think `config set` command should be restricted only for setting `token` for now

myishay (commented):
looks cool! gave you 1 comment

cmd/config/set.go
dimabru: this is intentional. config set will be able to change any key-value in config.yaml file

PR #85, opened 2021-07-05
Dat 3240 support pipes

myishay:


cmd/test/main.go
```diff
@@ -106,6 +119,22 @@ func test(ctx *TestCommandContext, paths []string, flags TestCommandFlags) error
 		return err
 	}
 
+	if paths[0] == "-" {
+		if len(paths) > 1 {
+			return fmt.Errorf(fmt.Sprintf("Unexpected args: [%s]", strings.Join(paths[1:], ",")))
+		}
+		tempFile, err := ioutil.TempFile("", "datree_temp_*.yaml")
```
alexfedin, line 134: you should generate a new name for each run, maybe using the date, in case the process will die before deleting the temp file so the next run will not get affected

alexfedin (changes requested):

cmd/test/main.go
myishay: the `*` in the second argument does exactly that :)

cmd/test/main.go
```diff
@@ -67,28 +71,45 @@ func New(ctx *TestCommandContext) *cobra.Command {
 	testCommand := &cobra.Command{
 		Use:   "test",
 		Short: "Execute static analysis for pattern",
-		Long:  `Execute static analysis for pattern. Input should be glob`,
+		Long:  "Execute static analysis for pattern. Input should be glob or `-` for stdin",
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
```
dimabru, line 76: you can do instead:
`Args: cobra.MinimumNArgs(1)`

cmd/test/main.go
```diff
@@ -169,7 +202,6 @@ func test(ctx *TestCommandContext, paths []string, flags TestCommandFlags) error
 	if len(invalidK8sFiles) > 0 {
 		err = ctx.Evaluator.UpdateFailedK8sValidation(invalidK8sFiles, createEvaluationResponse.EvaluationId, stopEvaluation)
 		if err != nil {
-			fmt.Println(err.Error())
 			return err
```
dimabru, line 205: you removed the err print but didn't defer it in the current scope. what will happen with errors in this case?

dimabru (approved):

PR #94, opened 2021-07-11
feat: Policy flag

dimabru:


bl/evaluation/evaluator.go
```diff
@@ -34,10 +34,11 @@ type EvaluationResults struct {
 	}
 }
 
-func (e *Evaluator) CreateEvaluation(cliId string, cliVersion string, k8sVersion string) (*cliClient.CreateEvaluationResponse, error) {
+func (e *Evaluator) CreateEvaluation(cliId string, cliVersion string, k8sVersion string, policyName string) (*cliClient.CreateEvaluationResponse, error) {
```
noaabarki, line 37: I wonder if it's time to group some of the parameters, maybe something like `cliInfo, options` this will be much handy to send all the flags as options

bl/evaluation/evaluator_test.go
```diff
@@ -77,11 +77,12 @@ func TestCreateEvaluation(t *testing.T) {
 		cliId := "test_token"
 		cliVersion := "0.0.7"
 		k8sVersion := "1.18.1"
+		policyName := "Default"
 
 		mockedCliClient.On("CreateEvaluation", mock.Anything).Return(&cliClient.CreateEvaluationResponse{EvaluationId: 1, K8sVersion: k8sVersion}, nil)
 
 		expectedCreateEvaluationResponse := &cliClient.CreateEvaluationResponse{EvaluationId: 1, K8sVersion: k8sVersion}
-		createEvaluationResponse, _ := evaluator.CreateEvaluation(cliId, cliVersion, k8sVersion)
+		createEvaluationResponse, _ := evaluator.CreateEvaluation(cliId, cliVersion, k8sVersion, policyName)
```
noaabarki, line 85: I suggest adding tests for policy option

bl/evaluation/evaluator.go
```diff
@@ -34,10 +34,11 @@ type EvaluationResults struct {
...
 	createEvaluationResponse, err := e.cliClient.CreateEvaluation(&cliClient.CreateEvaluationRequest{
 		K8sVersion: &k8sVersion,
 		CliId:      cliId,
+		PolicyName: policyName,
```
noaabarki, line 41: What if the policy isn't provided?

noaabarki (changes requested):

bl/evaluation/evaluator.go
dimabru: policy has a default value. even if it is not provided, cobra will create a policy flag with value `Default`

bl/evaluation/evaluator.go
dimabru: I agree, but this can turn out into a heavy change so I think I'll leave it to a smaller pr :)

bl/evaluation/evaluator_test.go
dimabru: Adding a test to make sure policy name is set to correct value before `createEvaluation` step. Unfortunately, since the policy name check is performed in the backend, there's no way of testing bad policy names behavior

PR #111, opened 2021-08-01
Dat 3357 parse and print accurrences

royhadad:


pkg/printer/printer.go
```diff
@@ -189,3 +194,11 @@ func (p *Printer) printPassedYamlValidation() {
 func (p *Printer) printSkippedPolicyCheck() {
 	p.printInColor("[?] Policy check didn’t run for this file\n", p.theme.Colors.White)
 }
+
+func (p *Printer) getStringOrNotAvailable(str string) string {
```
dimabru, line 208: Nice!!

pkg/printer/printer.go
```diff
@@ -189,3 +194,11 @@ func (p *Printer) printPassedYamlValidation() {
...
+	if str == "" {
```
dimabru, line 209: Will this work if `str` is `undefined`?
I think we can address that as well

bl/evaluation/printer.go
```diff
@@ -129,8 +129,9 @@ func parseToPrinterWarnings(results *EvaluationResults, invalidYamlFiles []*vali
 				rule := rules[ruleId]
 				details := printer.WarningInfo{
 					Caption:     rule.Name,
-					Occurrences: rule.Count,
+					Occurrences: rule.getCount(),
```
myishay: Is the `getCount` available to you in this context? I guess not since it is not Capitalized

royhadad: somehow it still worked, weird

pkg/printer/printer.go
```diff
@@ -23,6 +24,7 @@ type WarningInfo struct {
 	Caption     string
 	Occurrences int
 	Suggestion  string
+	Matches     []*cliClient.Match
```
dimabru: I think `rule` object should not be aware of the concept of `matches`.
This is just the data type from the backend, but should not be part of the final object. I think saving a array of specific identifiers -> `metadataName` and `kind` is more accurate.

Matches includes rule-to-file mapping, while we are reversing the order of the print to file-to-rule mapping

dimabru (changes requested):

pkg/printer/printer.go
royhadad: in golang, string cannot be nil (only a pointer to a string can)

bl/evaluation/printer.go
royhadad: ohhh, I only used it in the "evaluation" package, so it still worked.
fixed it now.

dimabru (dismissed):

PR #129, opened 2021-08-09
fix: reset console colors back to default

sobi3ch:
At the moment when you install detree it will end up with [`\033[35m`](https://github.com/datreeio/datree/blob/staging/install.sh#L39) this color. This fix should work on both Linux and macOS and put the console colors back to default.


myishay:
LGTM, thanks!

myishay (approved):

Issue #81, opened 2021-07-01
Unable to view results in an Azure DevOps pipeline


grafikkaos:
**Is your feature request related to a problem? Please describe.**
There are limited output options for Datree.  When used with Azure DevOps, it would be nice to able to pulish the results of these checks

**Describe the solution you'd like**
A JUnit output format would be extremely welcome to make it easier to export the results.  This can then be viewed in the Azure DevOps pipeline.

eyarz:
I researched a little, and it looks like JUnit output for the Azure DevOps pipeline is just a regular XML output format.
Do you know what the structure of this XML should look like for it to be parsed automatically by Azure DevOps pipeline?

eyarz:
done.

PR #135, opened 2021-08-15
Add XML output support

guyzyl:
Relates to #81.
Doesn't completely solve it since Azure DevOps expects the XML to be in JUnit format, but it takes the project one step closer towards that goal.

### DIFF
--- a/bl/evaluation/evaluator.go
+++ b/bl/evaluation/evaluator.go
@@ -25,8 +25,10 @@ func New(c CLIClient) *Evaluator {
 	}
 }
 
+type FileNameRuleMapper map[string]map[int]*Rule 
+
 type EvaluationResults struct {
-	FileNameRuleMapper map[string]map[int]*Rule
+	FileNameRuleMapper FileNameRuleMapper
 	Summary            struct {
 		TotalFailedRules int
 		FilesCount       int
--- a/bl/evaluation/printer.go
+++ b/bl/evaluation/printer.go
@@ -2,11 +2,13 @@ package evaluation
 
 import (
 	"encoding/json"
+	"encoding/xml"
 	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
 	"strings"
+	"strconv"
 
 	"github.com/datreeio/datree/bl/validation"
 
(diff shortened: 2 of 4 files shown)


myishay (commented):
Thank you for your contribution!

I played a bit with the xml output and it seems like the items in `FileNameRuleMapper` are not applying xml format rules. 

I created a basic policy check to our `k8s-demo.yaml` file using only `Ensure workload has valid label values` rule and this was the xml output result:
```
<?xml version="1.0" encoding="UTF-8"?>
<FormattedOutput>
	<EvaluationResults>
		<FileNameRuleMapper>
			<./internal/fixtures/kube/k8s-demo.yaml>
				<9>
					<Rule>
						<ID>9</ID>
						<Name>Ensure workload has valid label values</Name>
						<FailSuggestion>Incorrect value for key(s) under `labels` - the vales syntax is not valid so the Kubernetes engine will not accept it</FailSuggestion>
						<OccurrencesDetails>
							<MetadataName>rss-site</MetadataName>
							<Kind>Deployment</Kind>
						</OccurrencesDetails>
					</Rule>
					<Rule>
						<ID>9</ID>
						<Name>Ensure workload has valid label values</Name>
						<FailSuggestion>Incorrect value for key(s) under `labels` - the vales syntax is not valid so the Kubernetes engine will not accept it</FailSuggestion>
						<OccurrencesDetails>
							<MetadataName>rss-site</MetadataName>
							<Kind>Deployment</Kind>
						</OccurrencesDetails>
					</Rule>
				</9>
			</./internal/fixtures/kube/k8s-demo.yaml>
		</FileNameRuleMapper>
		<Summary>
			<TotalFailedRules>1</TotalFailedRules>
			<FilesCount>1</FilesCount>
			<TotalPassedCount>0</TotalPassedCount>
		</Summary>
	</EvaluationResults>
	<EvaluationSummary>
		<ConfigsCount>1</ConfigsCount>
		<RulesCount>15</RulesCount>
		<FilesCount>1</FilesCount>
		<PassedYamlValidationCount>1</PassedYamlValidationCount>
		<PassedK8sValidationCount>1</PassedK8sValidationCount>
		<PassedPolicyCheckCount>0</PassedPolicyCheckCount>
	</EvaluationSummary>
</FormattedOutput>
```

A few things that I find a bit problematic:
- xml tags should not contain `/` character as the tag type, as in the filename tag for ecample: `<./internal/fixtures/kube/k8s-demo.yaml>`. A good solution would be changing the tag to something like `<File filename="./internal/fixtures/kube/k8s-demo.yaml">`
- xml tags should not start with a number, as in rule id tag for example: `<9>`. A good solutions would be remving this tag since it encapsulates a `<Rule>` tag anyway.


Also, for some reason I didn't check yet, I see 2 appearances of rule id 9 in the tested file, whereas only 1 appearance should be the correct output (you can look at the yaml/json/text output to see how it should look like). Seems like a bug to me 🐛.

@guyzyl  WDYT?

guyzyl:
@myishay I totally agree with all of your comments and I pushed fixed for all of them:
- Removed the rule integer id tag (ie: <9>)
- Changed `filepath` to be an attribute inside a `File` tag
- Fixed issue where the rule is outputted twice
- Improved var naming on the way :)

myishay (approved):
I'm good with the changes, and I also liked the naming improvements!

PR #142, opened 2021-08-22
fix: build in Dockerfile 

dimabru:


Dockerfile
```diff
@@ -1,7 +1,12 @@
...
-WORKDIR /app
+WORKDIR /go/src/app
+COPY . .
 RUN apt-get update && apt-get install -y curl grep unzip bash
 RUN curl https://get.datree.io | /bin/bash
 
-ENTRYPOINT ["datree"]
+RUN go get -d -v ./...
+RUN go build -tags main -ldflags="-X github.com/datreeio/datree/cmd.CliVersion=___CLI_VERSION" -v ./...
+RUN go install -v ./...
```
myishay, line 10: `./...` doesn't match our makefile, please align with the makefile (either adding to makefile or removing here)

myishay (dismissed):

Dockerfile
dimabru: removed

PR #143, opened 2021-08-22
feat: Dat 3375 publish command

royhadad:


pkg/cliClient/cliClient_test.go
```diff
@@ -174,6 +199,30 @@ func TestGetVersionMessage(t *testing.T) {
 	}
 }
 
+func TestPublishPolicies(t *testing.T) {
+	tests := []*PublishPoliciesTestCase{
+		test_publishPolicies_success(),
```
dimabru, line 206: add a failed test. we need to make sure we can handle errors from the backend, this should happen a lot

dimabru (dismissed):

PR #157, opened 2021-09-01
fix: remove commits by travis during builds

dimabru:


Dockerfile
```diff
@@ -1,10 +1,12 @@
 FROM golang:1.15-alpine
+RUN apk --no-cache add curl
 
 WORKDIR /go/src/app
 COPY . .
 
+RUN curl --silent "https://api.github.com/repos/datreeio/datree/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/' > cli-version
 RUN go get -d -v ./...
```
myishay, line 8: I like it!
I think it will be a better idea saving the version as a variable rather than as a file.
something like `cli_version=$(curl --silent "https://api.github.com/repos/datreeio/datree/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')`

dimabru: not possible. docker evaluates each layer seperately and saves image state when moving from one command to another. there's no memory sharing

internal/release/brew_formula_generator.sh
```diff
@@ -0,0 +1,45 @@
...
+  then
+    echo "Not enough arguments supplied"
+    exit
+fi
+
+VERSION=$1
+DESTINATION_FOLDER=$2
+
+cat ./dist/checksums.txt
+
```
myishay, line 13: I guess this is redundant

internal/release/brew_formula_generator.sh
```diff
@@ -0,0 +1,45 @@
...
+if [ $# -lt 2 ]
+  then
+    echo "Not enough arguments supplied"
+    exit
+fi
+
+VERSION=$1
+DESTINATION_FOLDER=$2
+
+cat ./dist/checksums.txt
```
myishay: I guess this is redundant

dimabru: removed

alexfedin (approved):

PR #181, opened 2021-09-29
fix: prompt message color + error haneling

myishay:


cmd/test/main.go
```diff
@@ -261,11 +262,18 @@ func test(ctx *TestCommandContext, paths []string, flags TestCommandFlags) error
 	err = evaluation.PrintResults(results, invalidYamlFiles, invalidK8sFiles, evaluationSummary, fmt.Sprintf("https://app.datree.io/login?cliId=%s", localConfigContent.CliId), flags.Output, ctx.Printer, createEvaluationResponse.K8sVersion, createEvaluationResponse.PolicyName)
 
 	if len(createEvaluationResponse.PromptMessage) > 0 {
-		fmt.Println(createEvaluationResponse.PromptMessage + " (Y/n)")
-		answer, _, _ := keyboard.GetSingleKey()
+		fmt.Fprint(color.Output, color.HiCyanString("\n\n"+createEvaluationResponse.PromptMessage+" (Y/n)\n"))
```
noaabarki: I think maybe we should wrap it in private functions so it can be more readable

noaabarki (changes requested):

noaabarki (dismissed):

PR #198, opened 2021-10-12
feat: --only-k8s-files flag

### LINKED ISSUE #187: Non strict mode - ignore files are not k8s files

when a user has several files (e.g. "user_configs.yaml") in a dir that not all of them are K8s files and he runs datree on the entire dir content, he will get errors (not valid yaml/k8s file check will fail) on those files.

add a flag that will ignore files that are not K8s files.
we can do that by checking for mandatory properties such as "apiVersion" or "kind".

dimabru:


bl/validation/k8sValidator.go
```diff
@@ -62,7 +71,20 @@ func (val *K8sValidator) ValidateResources(filesConfigurationsChan chan *extract
 			}
 		}
 	}()
-	return validK8sFilesConfigurationsChan, invalidK8sFilesChan
+	return validK8sFilesConfigurationsChan, invalidK8sFilesChan, ignoredFilesChan
+}
+
+func (val *K8sValidator) validatePotentialK8sFile(fileConfigurations []extractor.Configuration) bool {
+	for _, configuration := range fileConfigurations {
```
alexfedin, line 92: I wouldn't call it validation, this is more of a classification, isK8sFile would be a better name

bl/validation/k8sValidator.go
```diff
@@ -33,17 +33,26 @@ func (val *K8sValidator) InitClient(k8sVersion string, ignoreMissingSchemas bool
...
 
 	go func() {
 		defer func() {
 			close(invalidK8sFilesChan)
 			close(validK8sFilesConfigurationsChan)
+			close(ignoredFilesChan)
 		}()
 
 		for fileConfigurations := range filesConfigurationsChan {
+			if onlyK8sFiles {
```
alexfedin: This way the k8s classification and validation are tied together and will make changes for either of them potentially affect each other

The classification of the k8s files should happen outside of this function. The validation function should only receive k8s files and decide if they are valid k8s files or not.

alexfedin (changes requested):

PR #207, opened 2021-10-18
feat: create release-candidate on every push to main

dimabru:


internal/release/deploy_release_candidate.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ "$latestRcTag" == "" ]; then
+    nextVersion=$MAJOR_VERSION.$MINOR_VERSION.0
+else
+    nextVersion=$(echo $latestRcTag | awk -F. '{$NF = $NF + 1;} 1' | sed 's/ /./g')
+fi
+
+export DATREE_BUILD_VERSION=$nextVersion-rc
+echo $DATREE_BUILD_VERSION
+
```
noaabarki: This is one awesome-written script!
However, I think we should move it to `scripts` folder outside of internal.

dimabru: done

PR #214, opened 2021-10-20
fix: handle empty files as errors (#213)

### LINKED ISSUE #213: Empty files pass Kubernetes validation
**Describe the bug**
Running datree test on empty files shows `invalid memory address or nil pointer dereference` exception.

**To Reproduce**
1. Run command 'test' on an empty file.

**Expected behavior**
Empty files should not pass the Kubernetes validation step.

**Screenshots**
image

 - OS: macOS

**Datree version (run `datree verion`):**
 - Version: 0.14.4

tipok:
Here is a simple fix handling empty files as errors instead of failing with NPE

### DIFF
--- a/bl/validation/k8sValidator.go
+++ b/bl/validation/k8sValidator.go
@@ -106,8 +106,16 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, error
 	if err != nil {
 		return false, []error{}, fmt.Errorf("failed opening %s: %s", filepath, &InvalidK8sSchemaError{ErrorMessage: err.Error()})
 	}
+	defer f.Close()
 
 	results := val.validationClient.Validate(filepath, f)
+
+	// Return an error if no valid configurations found
+	// Empty files are throwing errors in k8s
+	if len(results) == 1 && results[0].Status == kubeconformValidator.Empty{
+		return false, []error{&InvalidK8sSchemaError{ErrorMessage: "empty file"}}, nil
+	}
+
 	isValid := true
 	var validationErrors []error
 	for _, res := range results {
--- a/bl/validation/k8sValidator_test.go
+++ b/bl/validation/k8sValidator_test.go
@@ -24,6 +24,7 @@ func TestValidateResources(t *testing.T) {
 	test_valid_multiple_configurations(t)
 	test_valid_multiple_configurations_only_k8s_files(t)
 	test_invalid_file(t)
+	t.Run("test empty file", test_empty_file)
 }
 
 func test_valid_multiple_configurations(t *testing.T) {
@@ -84,6 +85,30 @@ func test_invalid_file(t *testing.T) {
 	}
 
 	path := "../../internal/fixtures/kube/invalidK8sSchema.yaml"
+
+	filesConfigurationsChan := make(chan *extractor.FileConfigurations, 1)
+	filesConfigurationsChan <- &extractor.FileConfigurations{
+		FileName:       path,
+		Configurations: []extractor.Configuration{},
+	}
+	close(filesConfigurationsChan)
+	_, invalidFilesChan := k8sValidator.ValidateResources(filesConfigurationsChan, 1)
+
+	for p := range invalidFilesChan {
+		assert.Equal(t, path, p.Path)
+	}
+}
+
+func test_empty_file(t *testing.T) {
+	validationClient := &mockValidationClient{}
+	validationClient.On("Validate", mock.Anything, mock.Anything).Return([]kubeconformValidator.Result{
+		{Status: kubeconformValidator.Invalid, Err: fmt.Errorf("empty file")},
+	})
+	k8sValidator := K8sValidator{
+		validationClient: validationClient,
+	}
+
+	path := "../../internal/fixtures/kube/empty.yaml"
 
 	filesConfigurationsChan := make(chan *extractor.FileConfigurations, 1)
 	filesConfigurationsChan <- &extractor.FileConfigurations{


dimabru:
Hi @tipok, Thanks for the pr!

I think this issue can be fixed a bit differently.
Datree has 3 validation steps. yaml validation, k8s validation, and finally rules evaluation.
Your pr fails empty yaml as part of the first yaml validation step. Empty yaml files are in fact valid yaml files. You can try validating an empty file in an online validator and see that it passes the validation. Also, our code tries to convert yaml to json and if it fails we throw an error. Clearly it passes since no errors are thrown.

As described in the attached issue, empty yaml files should fail during the second step, k8s validation, since empty yaml files are not valid k8s files.
In the second step, we are using a great open source k8s validator - [kubeconform](https://github.com/yannh/kubeconform)

It seems to me that the solution should focus on passing correct value to kubeconform to allow it to fail, or in case the bug is on kubeconform, open a pr there that fixes empty files validations

tipok:
Hi @dimabru, thanks for verifying.

Sounds legit, I'll move my fix to the suggested validator today or tomorrow.

tipok:
I found the code part with special handling built as a work around leading to the issue. 

```go
for _, res := range results {

    // A file might contain multiple resources
    // File starts with ---, the parser assumes a first empty resource
    if res.Status == kubeconformValidator.Invalid || res.Status == kubeconformValidator.Error {
        isValid = false
        validationErrors = append(validationErrors, &InvalidK8sSchemaError{ErrorMessage: res.Err.Error()})
    }
}
```

Not 100% sure if this behaviour should be handled as an issue in kubeconform or in Datree.
Do you have an opinion if this issue should be fixed in Datree or kubeconform?

For Datree I would add a check if any valid configuration is found otherwise return an error.

dimabru:
Yeah I think you're right. Great catch!
I think checking for kubeconform's behavior can be a good approach. I would try and run an empty file on kubeconform cli to see how it behaves. If it passes, open an issue in kubeconform because empty yaml files and not valid k8s files.

Regardless, we can certainly add a check on Datree side as you suggested. 

Should you decide to open an issue on kubeconform, can you please add a link to this pr so we can track it?

tipok:
I've changed the PR corresponding to our discussion. Besides of that I've created an issue for kubeconform, but I'm not sure if they are going to fix it in the library or the tool itself, current API implies the fix in the tool itself not the library.

[kubeconform/#80](https://github.com/yannh/kubeconform/issues/80)

yannh:
Hi @tipok  :wave:  Maintainer of Kubeconform here, just asking for some clarification since the devil lies in the detail :P 
Do you think kubeconform should fail when there is an empty  file, or an empty resource? Eg:
 * What if the file is not empty but  contains one, or multiple, yaml resources, like:

---
---
---

 * What if the file contains multiple resources, but several empty ones in the middle:

---
[one resource]
---
---
---
[another resource]

I haven't tried to check precisely which ones actually trigger errors in Kubernetes. So should Kubeconform fail on:
 * Empty files, or:
 * Files that contain no non-empty yaml document, or:
 * Files that contain no valid Kubernetes resources, or:
 * Files that contain any empty yaml document?

Thanks! And thanks for reporting this upstream to get my attention :rocket:

yannh:
Btw if any datree dev reads this - I didn't know datree's relies on Kubeconform's API! That's great :1st_place_medal:  I don't know if many people rely on it, and it's a very early draft. I would love feedback, whether the API works well, or whether I could make it work better. (referring to this https://github.com/yannh/kubeconform/blob/master/examples/main.go being used here https://github.com/datreeio/datree/blob/main/bl/validation/k8sValidator.go  ) 
:+1:

tipok:
@yannh Thank you for the response. I'm not 100% sure, to be honest. I guess @noaabarki is the person knowing best, because she opened the initial issue.

dimabru:
Hi @yannh, welcome to our repo :)
Your api proved to be very easy and intuitive to use, thanks for that 😄

Regarding empty yaml files, I'd love to get your input. 
I think the behavior should differ between files with two `---` seperators with nothing in between and empty files.
Helm users can easily output a file with empty configuration, but running empty files is not a reasonable k8s user behavior, and should be considered as bad file input (IMO, please feel free to change my mind).

Another approach can be - treat files as kubectl does, which in other words means: 
_if it's good for kubectl - it's good enough for me_

I personally prefer the first approach, but would love to know what you think.

yannh:
So - if the file is completely empty (not even an empty yaml document) - we should exit with -1?
Would love some suggestions on what you think the output of kubeconform should be, currently it is very "resources oriented" - we report on number of resources found, and whether they are valid or not.

bl/validation/k8sValidator.go
```diff
@@ -118,6 +122,20 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, error
...
+		}
+	}
+
+	// Return an error if no valid configurations found
+	// Empty files are throwing errors in k8s
+	if !seenValid && isValid {
+		isValid = false
+		validationErrors = append(
+			validationErrors,
+			&InvalidK8sSchemaError{ErrorMessage: fmt.Sprintf("empty file: %s", f.Name())},
```
dimabru: No need to print out the file name, it is already in context and is part of the output

bl/validation/k8sValidator.go
```diff
@@ -106,9 +106,13 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, error
 	if err != nil {
 		return false, []error{}, fmt.Errorf("failed opening %s: %s", filepath, &InvalidK8sSchemaError{ErrorMessage: err.Error()})
 	}
+	defer func() {
+		_ = f.Close()
+	}()
 
 	results := val.validationClient.Validate(filepath, f)
 	isValid := true
+	seenValid := false
```
dimabru: Adding a specific parameter that is set to true if a file contains at least one object feels redundant. I think you can simplify it. 
You can simply assert an error if `len(results) == 0` and add the same logic as on line 137 without all the conditional checks.

Also, this will return an error for files containing multiple `---` seperators with nothing in between (which is not an empty file).

dimabru (changes requested):

bl/validation/k8sValidator.go
tipok: Hi,

Thank you for verifying. 

1. Unfortunately `len(results) == 0` won't ever be true, because it's always having one result with the state of empty.
2. I'm not 100% sure if this should be the case, I'm not seeing a difference between an absolutely empty file and a file containing separators only, which I would think of as an user error.

bl/validation/k8sValidator.go
```diff
@@ -106,9 +106,13 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, error
 	if err != nil {
 		return false, []error{}, fmt.Errorf("failed opening %s: %s", filepath, &InvalidK8sSchemaError{ErrorMessage: err.Error()})
 	}
+	defer func() {
+		_ = f.Close()
```
noaabarki: There's no need to wrap the defer inside a function, simply use `defer f.Close()`. In addition, we don't want to ignore errors so if for any reason closing the file failed we would want to return the error.

bl/validation/k8sValidator.go
dimabru: Yeah sorry, missed it. `len(results) == 1 && results[0].Status == kubeconformValidator.Empty` should be enough.

Regarding [2], files with seperators can be in fact valid `helm template` outputs. I agree its not very common, but users can run conditional helm charts that end up as empty resources inside a single k8s yaml file. Not to say we won't change that in the future, but as long as nobody has any issues with it we prefer no to change a working behavior.

tipok: OK, thank you for explaining. I'll update the PR.

dimabru:
> So - if the file is completely empty (not even an empty yaml document) - we should exit with -1? Would love some suggestions on what you think the output of kubeconform should be, currently it is very "resources oriented" - we report on number of resources found, and whether they are valid or not.

I guess if no users are complaining about it you can keep kubeconforms validation the way it is right now. 
I personally think that from the k8s user's perspective, there's no need for them to add empty k8s files, and if `kubeconform` purpose is to shift-left the validation process, then it should alert users trying to validate empty files.
Regardless, we're going to add that behavior in `datree` cli

noaabarki (changes requested):

yannh:
@dimabru :+1:  I'd be pretty happy to change it, just trying to understand what actually fails with Kubernetes, whether its only completely empty files (which wouldn't be too hard to do), or more, since the implementation would be slightly different :) 

Note: given the current API, I think this PR is probably the best way forward :bow:

bl/validation/k8sValidator.go
tipok: As of my understanding the way I defined the defer function is behaving exactly the same as
`defer f.Close()` which is ignoring return values.
If there is need of returning the error it has to be done explicitly, let me know if I should refactor the code accordingly.

bl/validation/k8sValidator.go
```diff
@@ -120,6 +123,16 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, error
 		}
 	}
 
+	// Return an error if no valid configurations found
+	// Empty files are throwing errors in k8s
+	if isValid && (len(results) == 1 && results[0].Status == kubeconformValidator.Empty){
```
dimabru: I think this piece of code is redundant. You can do the check before going inside the loop.
Something like this
```go
if len(results) == 1 && results[0].Status == kubeconformValidator.Empty {
  return false, []error{&InvalidK8sSchemaError{ErrorMessage: "empty file"}}, nil
}
```
Right after line 113

dimabru (changes requested):

bl/validation/k8sValidator.go
tipok: you're totally right

dimabru (dismissed):

dimabru:
Great, thanks for the effort. Please rebase from staging as the branch is out of date, and wait for @noaabarki to approve as well

bl/validation/k8sValidator.go
noaabarki: Hey @tipok! Haha, you are absolutely correct, I completely missed it, sorry for the trouble! We decided we want to ignore the error, so simply use `defer f.Close()`

tipok: ok, I'll change it, thank you. Just my 2 cents on that topic: 
1. Some linter would show issues on that because errors are not handled correctly 
2. As discovered in that thread it's not that obvious what happens with returned values on defer if not handled explicitly (as in my solution)

 Therefore I personally do prefer handling it explicitly so it's totally clear what is happening.

noaabarki: Thank you! I understand what you mean. However, we prefer to continue with this until we'll gain more knowledge about how we would want to handle it.

noaabarki (approved):

dimabru (approved):

bluebrown:
Why is an empty file an error? This does not make sense to me and its makes it more fiddley to use as git pre commit hook.

eyarz:
The logic behind this decision is that maybe you don't intend to have an empty K8s file, so you want to be notified about that. Our community decided that this is the expected behavior. 

@bluebrown, you can use the `--only-k8s-files` flag to filter out empty files.
Don't forget to also use the glob pattern to only target `yaml` files - `$datree test [path]/*.yaml --only-k8s-files`

bluebrown:
@eyarz , Thanks for the hint.  I am using this flag. Now I am not able to reproduce it any more. It was when piping the output from git diff into a docker container. Datree would see some tmp yaml file and complain that it's empty.   

This can also be solved reliably using `xargs --no-run-if-empty`.

bluebrown:
@eyarz, It appears that `--only-k8s-files` has no effect when providing all files explicitly. For example, via find and xargs. Or am I doing something wrong?

hadar-co:
> @eyarz, It appears that `--only-k8s-files` has no effect when providing all files explicitly. For example, via find and xargs. Or am I doing something wrong?

Hi @bluebrown can you provide an example command? I'm not sure I understand what you mean.

romanlab:
@bluebrown i wanted to make sure that the flag works and it seems to ignore non k8s files even when passed explicitly, but even then it fails with empty files. @hadar-co try running `datree test --only-k8s-files` on an empty file and on a yaml file that's not empty but neither a k8s file and you'll see that it fails for empty but passes for non empty

bluebrown:
It throws errors on me if there is a Kptfile. This file might come from a bunch of file names using xargs. 

```
$ datree test --only-k8s-files Kptfile
>>  File: /home/blue/projects/ns/kube/gitops/sources/services/kpt-sample/Kptfile

[V] YAML validation
[X] Kubernetes schema validation

❌  k8s schema validation error: could not find schema for Kptfile
You can skip files with missing schemas instead of failing by using the `--ignore-missing-schemas` flag


[?] Policy check didn't run for this file


(Summary)

- Passing YAML validation: 1/1

- Passing Kubernetes (1.20.0) schema validation: 0/1

- Passing policy check: 0/1

+-----------------------------------+------------------------------------------------------+
| Enabled rules in policy "Default" | 21                                                   |
| Configs tested against policy     | 0                                                    |
| Total rules evaluated             | 0                                                    |
| Total rules skipped               | 0                                                    |
| Total rules failed                | 0                                                    |
| Total rules passed                | 0                                                    |
| See all rules in policy           | https://app.datree.io/login?t=e4fatJVcLeR1148RvK82C9 |
+-----------------------------------+------------------------------------------------------+
```

The file looks alot like a kubernetes file, maybe thats why.

```
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: test
  annotations:
    config.kubernetes.io/local-config: "true"
info:
  description: datree test
```

The exit code is 2. So my pre-commit hook fails.

```bash
$ echo $?
2
```

bluebrown:
I guess this is the wrong issue for this. Perhaps I should start a discussion or open a new issue.

hadar-co:
I believe the fix should be that the `--only-k8s-files` flag should ignore empty files as well. @bluebrown can you open a new issue for that?

hadar-co:
@bluebrown I opened an issue - https://github.com/datreeio/datree/issues/810

PR #216, opened 2021-10-20
fix: create release workflow

myishay:


.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
+name: Release Workflow
+on:
+  workflow_dispatch:
+    inputs:
+      name:
+        description: "Triggered by:"
```
dimabru, line 6: will github provide us info on who triggered the release without this input?

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+        required: true
+        default: "myishay"
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
```
dimabru, line 13: the step that triggers the travis build is missing here

.github/workflows/release.yml
myishay: I didn't  find how to do it yet, but I will keep searching

myishay: found it! resolving this conversation

.github/workflows/release.yml
myishay: added it now

myishay: @dimabru lmk what you think

dimabru (dismissed):

PR #218, opened 2021-10-20
Test refactor

alexfedin:
test command refactor:
1. introduce new evaluationManager entity to store files validations in one place so the evaluation logic can be extracted from the test command (still a long way to go to break this piece of code into smaller functions, but this will allow us to move forward)
2.  extract evaluate part from the test function
3. defer the close spinner functionality for clarity that the spinner should spin only during the evaluation


cmd/test/validationManager.go
```diff
@@ -0,0 +1,80 @@
...
+	ignoredFiles                []extractor.FileConfigurations
+}
+
+func (v *ValidationManager) AggregateInvalidYamlFiles(invalidFilesChan chan *validation.InvalidYamlFile) {
+	for invalidFile := range invalidFilesChan {
+		v.invalidYamlFiles = append(v.invalidYamlFiles, invalidFile)
+	}
+}
+
+func (v *ValidationManager) InvalidYamlFiles() []*validation.InvalidYamlFile {
```
myishay, line 21: I think we should name each of these getters with prefix "Get"

alexfedin: from the go docs: https://golang.org/doc/effective_go#Getters
"it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner."

myishay (approved):

PR #234, opened 2021-10-25
fix: add manually triggered release

myishay:


scripts/release.sh
```diff
@@ -1,3 +1,33 @@
...
+
+if [ $latestRcTag == "" ]; then
+    echo "couldn't find latestRcTag"
+    exit 1
+fi
+
+git checkout $latestRcTag
+
+release_tag=${latestRcTag%-rc}
+git tag $release_tag -a -m "Generated tag from manual TravisCI for build $TRAVIS_BUILD_NUMBER"
```
dimabru: add `to production`

scripts/release.sh
```diff
@@ -1,3 +1,33 @@
...
+git push --tags
+
+# Secure private key
```
dimabru: we have a script for that
`sign_application.sh`

dimabru:
delete `on: main` condition from old deploy phase in travis

PR #236, opened 2021-10-26
feat: use stale action to mark stale issues

### LINKED ISSUE #231: Replace "stale app" with the "stale Action"
Today, we are using the outdated [stale app](https://github.com/datreeio/datree/blob/staging/.github/stale.yml) to close stale issues

use the [stale Action](https://github.com/actions/stale) that has more capabilities

**Additional context**
we should also configure the stale app to close stale PR

prasadkatti:
Fixes: #231


.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Close stale issues'
+on:
+  schedule:
+    - cron: '0 */6 * * *'
```
eyarz: In all the [examples](https://github.com/actions/stale#usage), they (github) run the cron job once a day (`30 1 * * *`). Do you have a reason why you decided to set it to every 6h?

eyarz (changes requested):

.github/workflows/stale.yml
prasadkatti: No particular reason. Just thought it would be better to run it more frequently than once a day. But I will stick to the convention and change it back.

romanlab (dismissed):

eyarz (dismissed):

dimabru:
Thanks!!

eyarz (approved):

PR #237, opened 2021-10-26
feat: Command auto completion option to terminal datreeio#183

### LINKED ISSUE #183: Add auto completion to terminal
When using datree cli, there's no support for auto complete for next possible command 

Add auto-completion support for terminal (zsh, bashrc, etc..)
Can be part of the installation flow for datree

imrushi:
Added the command auto-completion like datree [tab][tab] it will give all following commands version, completion, config, test...

```
$ datree [tab][tab]
completion  config      help        publish     test        version
```
To test follow the below steps:
1. compile code with `go build -o datree main.go`
2. `./datree completion -h` we will get a description of how to use the completion command for [bash|zsh|fish|powershell].
3. For example bash we will use 
`$ datree completion bash > /etc/bash_completion.d/datree`
`$ source /etc/bash_completion.d/datree`
4. Now you will get auto-completion of command like below
```
$ datree [tab][tab]
completion  config      help        publish     test        version
$ datree completion [tab][tab]
bash        fish        powershell  zsh
```


cmd/completion/main.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+type Messager interface {
+	LoadVersionMessages(messages chan *messager.VersionMessage, cliVersion string)
+}
+
+type Printer interface {
+	PrintMessage(messageText string, messageColor string)
+}
+type CompletionCommandContext struct {
```
dimabru: Unlike other commands, `completion` has no dependenies on external logic. Since you're not using any of the completion command context paramteres, you can delete the type itself

cmd/completion/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	b := bytes.NewBufferString("bash") // Same for zsh, fish, powershell
+	cmd.SetOut(b)
+	cmd.SetArgs([]string{"bash"}) // Same for zsh, fish, powershell
+	cmd.Execute()
+	out, err := ioutil.ReadAll(b)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// repalce bash with zsh, fish, powershell as per var b & SetArgs
```
dimabru: I think the test should cover all potential flags. I removed the powershell flag and the test passed succesfully

cmd/completion/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+	if string(out) != "bash" {
```
dimabru: We should (at least partially), make sure the output matches the expected output of the completion command. Here you only check for the flag.
Maybe just check the first line of the output matches `# bash completion for datree` ?

dimabru (changes requested):
Thanks! This is a great pr!!
I made some comments, should be small and easy fixes

cmd/completion/main_test.go
imrushi: Could you please provide an example if you have any?
also could you please elaborate more `Maybe just check the first line of the output matches # bash completion for datree ?`

imrushi: oh, I got it.

dimabru: :-)

imrushi: Hey @dimabru, 
Can you please help me with how should I get the output of a command that is I am executing from the test file? I am so stuck at this point.

dimabru:
Thanks!!

imrushi:
Hi @dimabru,
Please review, all required fixes/changes are done.

cmd/completion/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+func Test_ExecuteCommand(t *testing.T) {
+
+	outBuff := new(bytes.Buffer)
+	var tt string
+	cmd := New()
+
+	b := bytes.NewBufferString("bash") // repalce bash with zsh, fish, powershell for different shell
+	cmd.SetOut(b)
+	cmd.SetArgs([]string{"bash"}) // repalce bash with zsh, fish, powershell for different shell
+	cmd.Run = func(cmd *cobra.Command, args []string) {
```
dimabru: The unit-tests role is to test the logic of the cmd. Here you override the logic we want to test. If I go to the completion cmd file and delete or modify all the code there, this test will still pass because it has nothing to do with the actual command.

I recommend having a look here: https://github.com/datreeio/datree/blob/main/pkg/printer/printer_test.go
we tested the printer package with override for the os.Stdout method. That way we're able to make sure the function outputs what we want.

Also, please make sure you cover all possible outputs

dimabru (changes requested):

imrushi:
As per your recommendation, I have made the required changes. Please have a look at it.
I have tried to cover all possible checks from my end.
Please review.

dimabru (approved):

PR #254, opened 2021-10-30
feat: os architecture support for i386 and arm

### LINKED ISSUE #203: Support installation for architectures other than x86_64
Installation script supports only x86_64 architecture. 386 and arm64 are not supported.

Detect the machine's architecture and download the relevant release in the install script

**Additional context**
the relevant file that should be changed should be `install.sh` and `windows_install.ps1`

amustaque97:
Fixes: #203 

This pr add support to install correct zip file based on OS architecture.

### Changes done in the PR
- Updated `install.sh` file to get architecture information and then grep the correct download link.
- Updated `windows_install.ps1` to get architecture info then proceed ahead with the installation.


myishay (changes requested):
@amustaque97  Thank you for the PR!
I tried your script on a Linux machine running on arm64 architecture and it errors :/

It is because the command `uname -m` outputs this: `aarch64`
and there are no releases that match the regex `"browser_download_url.*\_Linux_aarch64.zip")`

I guess you should find a better way to determine the machine arch in bash

myishay:
Also, I saw your PR is against staging branch. Staging branch is deprecated and will soon be deleted.
Thanks!!

amustaque97:
After spending some time on the internet. I realize there is no better way or to say error prone safe method to identify OS architecture. Since in ARM we have different architecture (full list can be seen here https://en.wikipedia.org/wiki/ARM_architecture)

- After going through stackoverflow link: https://stackoverflow.com/questions/48678152/how-to-detect-386-amd64-arm-or-arm64-os-architecture-via-shell-bash . Architecture will be different on different ARM devices. We are only supporting `arm64` in our case.

It would be great if you share how are you testing for ARM architecture?
In mac also it is a bit confusing as it shows both `x86_64 i386`
Screenshot 2021-10-31 at 10 19 09 PM

If this works fine for now. My opinion would be to consider this as of now. Will keep in mind to improve if we get some better solution in coming future.

What do you think?

install.sh
```diff
@@ -3,8 +3,13 @@
 set -e
 
 osName=$(uname -s)
-DOWNLOAD_URL=$(curl --silent "https://api.github.com/repos/datreeio/datree/releases/latest" | grep -o "browser_download_url.*\_${osName}_x86_64.zip")
 
+osArchitecture=$(uname -m)
+if [[ osArchitecture == *'aarch'* || osArchitecture == *'arm'* ]]; then
+	osArchitecture = 'arm64'
```
myishay: This if statement is not working well since you are using a string `osArchitecture` and not the actual variable `$osArchitecture `
also, you shouldn't have spaces in the assignment of `osArchitecture` to `arm64`

myishay (changes requested):
I think your suggestion to handle arm architecture are good enough for now.
Please make the desired changes and I think we are good to go.

install.sh
amustaque97: Thanks for pointing out the issues. I have addressed review comment.

myishay (approved):

PR #255, opened 2021-10-30
fix: show usage message with unrecognized flag

### LINKED ISSUE #244: CLI doesn't show Usage message with unrecognized flag
**Describe the bug**
When running `datree test --foobar` with an unrecognized flag `--foobar`, the cli doesn't show Usage message (whereas `datree version --foobar` does return Usage mesage

**To Reproduce**
1. Run command `datree test ~/.datree/k8s-demo.yaml --foobar`
2. nothing happens (only exit with nonzero)
3. Run command `datree version --foobar`
4. I can see Usage message

**Expected behavior**
should show Usage message
**Screenshots**
Screen Shot 2021-10-26 at 17 47 23

 - OS: macOS

**Datree version (run `datree verion`):**
 - Version: 0.14.17

amustaque97:
Fixes #244 

### Changes proposed in the PR
- Removed `SilenceUsage` and `SilenceErrors` from **test** command
- Removed `SilenceUsage` and `SilenceErrors` from **publish** command

### Screenshot
image


dimabru (changes requested):
Please note that our cli can fail with exit code 1 when rules are failing during standard execution. This is something that we don't want to happen when the cli errors on valid error.

This will print error and usage on valid output as well
For example (after running `make build` on your pr):
image

`Error` and `Usage` text is not required here
We need a solution that will print usage only in case the flag is unknown

amustaque97:
> Please note that our cli can fail with exit code 1 when rules are failing during standard execution. This is something that we don't want to happen when the cli errors on valid error.
> 
> This will print error and usage on valid output as well For example (after running `make build` on your pr): image
> 
> `Error` and `Usage` text is not required here We need a solution that will print usage only in case the flag is unknown

So shall I remove it from publish command? This is what you mean?

amustaque97:
I got it. what is the issue. Let me work on the fix. Since removing `SilenceUsage` and `SilenceErrors` has introduced some new bugs in this PR.

amustaque97:
Thanks for the review. I have addressed review comments. Tested both `datree test` and `datree publish` command thoroughly. Got to learn more about `datree`
Screenshot 2021-10-31 at 9 04 13 PM

myishay (approved):
LGTM

dimabru (approved):

dimabru:
Great PR :)

Issue #205, opened 2021-10-17
Automate contribution workflow


dimabru:
**Is your feature request related to a problem? Please describe.**
When we receive code contributions from the community, before reviewing the code logic we have to do a manual set of checks that can be automated

**Describe the solution you'd like**
Add a bot to check for incompatibility with our contribution guidelines before getting a real person to review the code

**Additional context**
Adding automated labels can also be a nice addition - `ltgm`, `cla-yes/no`, `fix-commit-message`, etc.

eyarz:
I will create a GitHub action workflow for that.
It will automatically verify the commit message and PR title.
@dimabru, anything else that you want it to do?

dimabru:
@eyarz no, thanks a lot

prasadkatti:
Hi. I would like to take a crack at this one.

eyarz:
@prasadkatti, go for it :)

PR #260, opened 2021-10-31
fix: check_commit_msg used for PRs titles only

### LINKED ISSUE #258: check_commit_msg workflow should ignore main branch
**Describe the bug**
The workflow for checking commit messages should not be executed after a pr is approved and pushed to main branch

kevholmes:
This resolves #258. When reviewing the original issue #205 it sounds like this was only supposed to run on `pull_request` events. My fix will ensure that this only runs when contributors submit a PR to the datree repo.


romanlab:
@kevholmes Actually, it would be great to give the contributor an indication before submitting a PR. It's ok that it runs on a `push` event, it just needs to ignore the `main` branch. If we do it this way, we can then refactor the script and use it with pre-commit.

maybe you can use 
```
on: 
  push:
    branches-ignore:
      - 'main'
``` 

to achieve that

WDYT?

kevholmes:
@romanlab I like the idea of adding the python script as a local pre-commit hook, that way the contributor could see the issue before they ever make a commit.

I'd say it's a pretty minute difference between `push` and `pull_request` really. If I fork datree's GH repo and push commits directly to my feature branch the `push` trigger _might_ run (as long as I have manually gone in and enabled Actions for my forked repo.) So it's not exactly automated for end users. I like reducing the scope of when this runs to something more narrow like `pull_request` as a good practice as the blanket on `push` could overlap with other git repo work you're doing that might not need to meet that Check's criteria and cause an unnecessary failure.

Also by default with on: `pull_request` that includes `synchronize` events (which are individual commits to a PR) so if a user makes a PR to this repo - but the check_commit_msg check fails - they can easily update the name of the commit in the GitHub UI to the correct (passing) format - and the check should re-run. If we included the `edited` event under a `pull_request` I believe we could catch the user updating the PR meta info itself and re-run the check - which I do not believe would be triggered on a `push` event since since that's outside of editing GH PRs (and their names.)

romanlab:
@kevholmes That's a valid point. But looking at the event structure for a `pull_request` event, i didn't see a `commits` key and not sure the script will work for the event. You can see the details [here](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request)

kevholmes:
> @kevholmes That's a valid point. But looking at the event structure for a `pull_request` event, i didn't see a `commits` key and not sure the script will work for the event. You can see the details [here](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request)

`pull_request` is a little confusing - they call the events `synchronize` when pushing follow-up commits to a PR. I think because of how GH handles PRs behind the scenes.

romanlab:
> > @kevholmes That's a valid point. But looking at the event structure for a `pull_request` event, i didn't see a `commits` key and not sure the script will work for the event. You can see the details [here](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request)
> 
> `pull_request` is a little confusing - they call the events `synchronize` when pushing follow-up commits to a PR. I think because of how GH handles PRs behind the scenes.

That's fine, i see the `synchronize` action but the event structure doesn't have a `commits` key. I approved the workflow, you can see the check failing

kevholmes:
Ahh ok, I see what you're saying @romanlab regarding the python [script](https://github.com/datreeio/datree/blob/main/scripts/check_commit_messages.py) possibly not working with this approach. It seems like we really want to run something that examines the GH PR's `title` and gives feedback regarding if they meet the standard or not. Each commit to a PR might not follow the approach, but ultimately its the PR `title` that will get squashed-and-merged into `main` and should follow the spec.

kevholmes:
Maybe using something like [this](https://github.com/marketplace/actions/pr-title-checker) to review the PR titles?

romanlab:
> Maybe using something like [this](https://github.com/marketplace/actions/pr-title-checker) to review the PR titles?

We prefer to not have external dependencies on projects with almost no community. `pr-title-checker` has only 31 stars. It'll be easier to adjust the script to validate the pr title.

kevholmes:
@romanlab How about this? I've refactored the python script and workflow to handle PR title checking instead of commit message checking. This has a more narrow scope and I think it meets the team's criteria. I have added some additional error handling to the python script as I noticed issues with errors when attempting to parse a `:` that might not be in the message if the author forgot it. I also added some additional verbosity to help explain to contributors why the check has failed (ie the message needs to be under 72 characters.)

I used some sample output from a test PR in my fork of datree's repo to test the Python script, here's the [sample json](https://gist.github.com/kevholmes/e9b18991c1be82535147a7695d4f2661). I am able to edit the PR title and verify the check will re-run at that time, and correctly passes or then fails depending on what I changed my PR title to.

kevholmes:
I've got some sample code running in my public fork if you'd like to review it there: https://github.com/kevholmes/datree/pull/2

.github/workflows/check_pr_title.yml
```diff
@@ -1,7 +1,14 @@
...
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+
 jobs:
-  check-commit:
+  check:
```
romanlab, line 11: Please change the script file name to the new name in the `run` command, the workflow fails at the moment.

romanlab (changes requested):

.github/workflows/check_pr_title.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repo content
+        uses: actions/checkout@v2
+      - name: setup python
+        uses: actions/setup-python@v2
+      - name: check PR title
+        run: python ./scripts/check_pr_title.py.py
```
kevholmes: 😆 I'll fix this now.

romanlab (approved):

PR #269, opened 2021-11-04
feat: add autocomplete usage instruction

### LINKED ISSUE #263: Autocompletion usage instructions in post install message
In order to have autocomplete for the CLI the user need to run the `completion` command

Print a message to the user that he should run the completion command if he wants autocompletion enabled after the installation is finished

**Additional context**
I suggest adding a short description about the completion command here 
image

amustaque97:
Fixes: #263 

### Changes proposed in the PR
- updated `install.sh` and `windows_install.ps1` to print `datree completion` command to enable shell completion.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -45,8 +45,9 @@ echo
 
 echo -e "\033[35m Usage: $ datree test ~/.datree/k8s-demo.yaml"
 
-echo -e " Using Helm? => https://hub.datree.io/helm-plugin \033[0m"
+echo -e " Using Helm? => https://hub.datree.io/helm-plugin"
 
+echo -e " Run 'datree completion $SHELL' to generate shell autocompletions \033[0m"
 tput init
 
 echo
--- a/install.sh
+++ b/install.sh
@@ -47,7 +47,7 @@ echo -e "\033[35m Usage: $ datree test ~/.datree/k8s-demo.yaml"
 
 echo -e " Using Helm? => https://hub.datree.io/helm-plugin"
 
-echo -e " Run 'datree completion $SHELL' to generate shell autocompletions \033[0m"
+echo -e " Run 'datree completion -h' to learn how to generate shell autocompletions \033[0m"
 tput init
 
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -45,8 +45,9 @@ echo
 
 echo -e "\033[35m Usage: $ datree test ~/.datree/k8s-demo.yaml"
 
-echo -e " Using Helm? => https://hub.datree.io/helm-plugin \033[0m"
+echo -e " Using Helm? => https://hub.datree.io/helm-plugin"
 
+echo -e " Run 'datree completion $SHELL' to generate shell autocompletions \033[0m"
```
dimabru: I use zsh as shell and running `echo $SHELL` outputs `/bin/zsh`
using that as an argument in the `completion` comment will not work because it expects `zsh` as value.

I think we can simply add a seperate print for each shell and let the user decide what he wants to do

dimabru: Also, running the command `datree completion $shell` will not result in adding auto completion. 
I suggest you run `datree completion -h` and see what needs to be done to add completion and try it yourself

dimabru (changes requested):

install.sh
amustaque97: Thanks for pointing this error. I configured autocompletion then I pushed changes but yeah wrote `$SHELL` keyword later on and pushed my changes without trying it. Since it is a echo command it didn't throw any error and I didn't noticed it.

> I think we can simply add a seperate print for each shell and let the user decide what he wants to do

In my opinion, we should just print `datree completion -h` only because adding each separate print will not look that good specially in case of bash i) linux and ii) macos. Moreover `datree completion -h` prints more structured output with some other info as compared to our own echo statements.

What do you think?

royhadad: Yep, I agree, just instruct the user to run "datree completion -h" to view to instructions

royhadad:
thank you for the contribution, good job! 😀

PR #311, opened 2021-12-11
[Hackathon] Security best practices policy

yummyweb:
## What
This PR adds an example of a security best practices policy. This policy checks the presence of certain fields like `runAsUser` and `runAsGroup`. Using this policy, DevOps engineers are able to handle permissions and accessibility of their clusters and containers.

## Why
This policy is based off of the Kubernetes documentation and certain articles of what constitutes as security best practices. One core aspect of Kubernetes security is `securityContext`. This PR is based on `securityContext` and allows Kubernetes engineers to handle permissions and accessibilities of their clusters and the containers within it. From the documentation, `securityContext` defines and manages privileges and access control settings.

### DIFF
--- a/examples/security-best-practices/README.md
+++ b/examples/security-best-practices/README.md
@@ -4,12 +4,14 @@ best practices such as setting user permissions and more. This policy requires t
 to be set to Pod, Deployment, StatefulSet or DaemonSet.
 
 __This policy helps to enforce the following security best practices:__
-* Ensuring the user permission is set
-* Ensuring the groups permission is set
-* Ensuring a privilege escalation flag is set
+* Ensure user permission is set
+* Ensure groups permission is set
+* Ensure privilege escalation is set
+* Ensure that filesystem accessibility is set
 
 ## Ensure user permission is set
-This flag controls which user ID the containers are run with. 
+This flag controls which user ID the containers are run with. If this value is 0, then
+the pod is run with non-root user.
 
 ### When this rule is failing?
 If the `runAsUser` key is missing from the securityContext section:  
@@ -20,7 +22,7 @@ spec:
     runAsUser: 1000
 ```
 
-__OR__ a invalid value of `runAsUser` is used:
+__OR__ an invalid value of `runAsUser` is used:
 ```
 kind: Pod
 spec:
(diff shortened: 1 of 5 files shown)


eyarz:
@yummyweb, this is an excellent idea for a policy! 
From what I see, your README file lists three rules, while your policy (`policy.yaml`) contains only one rule, is this on purpose?

yummyweb:
Yes, I was a bit apprehensive about that, let me change that.

AvinashNayak27:
@yummyweb  I think there is an Error in your Policy Schema,It is validating every input.Please Fix the Issue
Screenshot 2021-12-11 at 6 37 45 PM

yummyweb:
I believe the schema validator maybe incorrect since it is passing input for the sample policy. Please look at the pass.yaml and fail.yaml files I have provided and check using the cli.

CleanShot 2021-12-11 at 19 30 01@2x

eyarz:
@yummyweb why do you apply your policy only on resources type `Pod`?
It's also relevant for other types of resources (e.g. Deployment, DaemonSet, etc.).

The **yamlschemavalidator** is working as expected.
The issue is with your YAML Schema logic. 
Please review [Example #1](https://hub.datree.io/basic-examples#c6-example-1) for reference.

yummyweb:
Hi! My policy only applies to Pod since securityContext is set only for a Pod. It does not apply to other components (refer to Kubernetes documentation). The yaml schema validator does not seem to work since Example #1 also doesn't work. To test my policy, run it using the datree cli and test against fail.yaml.

yummyweb:
@eyarz The yamlschemavalidator does work, it turns out the input it was checked against was invalid, since it did not have the `metadata` or `spec` property, but my policy seems to work using the datree cli and the schema validator.

eyarz:
`securityContext` can be set for a Pod or **Container**.
Therefore, it can also be used with other resources:
https://stackoverflow.com/questions/57276023/setup-securitycontext-inside-kubernetes-deployment

yummyweb:
I have made the changes @eyarz

eyarz:

yummyweb:
@eyarz Why is my policy not moving forward?

PR #314, opened 2021-12-11
[Hackathon] Custom policy to check if configuration file consist the name of the deployment

kaiwalyakoparkar:
## What
While starting the configuration it's essential to make sure that there is `name` specified for the deployment.

## Why
Not having this might not raise errors by the native development kits (kubectl etc) but naming the deployment configurations while working with several clusters in a team is essential and cosidered a good practice.


eyarz:
@kaiwalyakoparkar it looks like your custom rule doesn't match the description in your policy README.

kaiwalyakoparkar:
@eyarz Sorry for that. I missed updating it as I was experimenting with lot of stuff. I will change it now.

eyarz:

PR #315, opened 2021-12-11
feat: add custom policy example (deployment_best_practices) 

AvinashNayak27:
## What
deployment_best_practices Consists of 2 policies One is `Rolling_Update_Deployment` and Other is `REVISION_HISTORY_LIMIT`
## Why
* `Rolling_Update_Deployment`  Ensures whether RollingUpdate deployment Strategy Enabled with `parameters` as `maxSurge`&`maxUnavailable`
* `REVISION_HISTORY_LIMIT` Helps in setting `revisionHistoryLimit` which helps K8's Admins to get History of Previous Deployments


eyarz:
@AvinashNayak27, this is a really nice use case to solve!
FYI, this PR contains one policy with two rules :)

I noticed that you didn't provide a `Fail.yaml` file, only `Pass.yaml`.
I recommend you to add it so you will not lose points on that.

AvinashNayak27:
Thank You For Your Reply.I wil be Adding `Fail.yaml` soon

eyarz (approved):

PR #316, opened 2021-12-11
[Hackathon] Policy to check Containers State

Shubham-com:
# What?

`Containers_best_practices` Policy check. These Container state is reported by Kubernetes
to check containers image state label during of the process of running the container .

# Why?

It helps to easily identify the Containers state label and  verify the correct state and have a valid state,
 A failed or terminated state gives an error.

lets say an example whenever a user try to run container as an image is still running the some kind of operations and requires more time to complete  then it undergoes in `waiting state`.While executing the containers image it is in `running State`.

If the state shows as a `terminated state` ,it a failed situation of running container as image.

### DIFF
--- a/examples/container policy check/README.md
+++ b/examples/container policy check/README.md
@@ -1,0 +1,46 @@
+# Policy: Containers_best_practices
+
+Pods are the smallest units of deployable ,which you can create and manage in Kubernetes.
+A Pod is a group of one or more containers, with shared storage and network resources, and 
+a specification  to run the containers. 
+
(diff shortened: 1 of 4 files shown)


eyarz:
@Shubham-com, I'm not sure I understand the use case.
The [Container state](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) is reported by Kubernetes and not by the user.
Can you please elaborate more about the use case that you're trying to solve?

Shubham-com:
Yes, I have added some description about it.

eyarz:

PR #320, opened 2021-12-12
feat: add custom policy example (security_checks)

viploveb:
This policy checks for container level security parameters like privilege escalation and root filesystem permissions. 
Developers and administrators usually skip over these parameters, since ignorance of these parameters can result in severe exploits therefore it is essential to check for them before deploying manifest files in production.
Due to their severity and importance these parameters are also a part of NSA Kubernetes Hardening Guide.

FYI - This isn't my only contribution to this hackathon, I am working on a project as well which uses datree. I will mention about this policy and my project in the devpost page.


eyarz:
@viploveb, good job!
I was hoping to see more rules because this is a good use case (NSA Kubernetes Hardening Guide) with potential ;)

viploveb:
@eyarz Thanks!
Do checkout my other project as well https://www.github.com/viploveb/shelm . It took sometime to create that and yaml schema was new to me as well 😅

eyarz (approved):

PR #322, opened 2021-12-12
feat: add custom policy example (horizontal_pod_autoscaling) 

theowlet:
## What
This policy is based on the HPA resource. It allows Kubernetes engineers to manage and configure HPA policies. The policy is created using a number of source like articles and documentation. They are documented below.

## Why
HPA is is ideal for scaling stateless applications. Using HPA with autoscaling can result in cost savings for clusters that see regular changes in demand by reducing the number of active nodes as the number of pods decreases. In order to manage this configuration, conventionally the `kubectl` command is used, but this policy can help in configuring HPA without the need of `kubectl`. This policy contains four rules for the HPA autoscaler: check for namespace, checking max replicas, checking min replicas and checking target cpu utilization.

## Sources
[Kubernetes Autoscaling](https://cloud.redhat.com/blog/kubernetes-autoscaling-3-common-methods-explained)


eyarz:
@theowlet how the rule [CUSTOM_WORKLOAD_NAMESPACE_CHECK](https://github.com/theowlet/datree/blob/main/examples/horizontal-pod-autoscaler-policy/policy.yaml#L14) is related to this use case?

theowlet:
@eyarz  Hi, this is required because according to [this](https://cloud.redhat.com/blog/kubernetes-autoscaling-3-common-methods-explained), namespace field is needed, hence the rule.

theowlet:
I have checked from the documentation and namespace is not required so I will remove the rule. Thank you for that feedback.

eyarz:
I also highly recommend you try and add more rules to your policy.
This is a good use case with potential ;)

theowlet:
@eyarz I have added a new rule.

eyarz (approved):

PR #325, opened 2021-12-12
feat: add custom policy example (security-policy)

Abhra303:
This PR contains a security related custom policy to let the kubernetes users follow security best practices. These policy rules are suggested by many blogs, conferences (e.g. cncf conference), cloud providers (e.g. oracle) and other organization including the CIS kubernetes benchmark.


yummyweb:
I see your policy only applies to Pods? Shouldn't these apply to other resources like Deployments, StatefulSets etc?

Abhra303:
> I see your policy only applies to Pods? Shouldn't these apply to other resources like Deployments, StatefulSets etc?

The last commit was pushed to address this. I added definitions in the schemas so that this work on all other resources you specified.
Thanks :)

eyarz (changes requested):
I noticed that your README file contains only 10 rules while your policy contains 12 rules.
Can you please update your README and fail/pass tests to include the missing rules?

Abhra303:
@eyarz wrote:

> I noticed that your README file contains only 10 rules while your policy contains 12 rules. Can you please update your README and fail/pass tests to include the missing rules?

Yeah sure! I was going to add those rules in the readme file but could not push them as hackathon was only few seconds left.

eyarz:
@Abhra303, please add it :)

Abhra303:
@eyarz done.

eyarz (approved):

PR #328, opened 2021-12-12
feat: add custom policy examples

dhanushm7:
This PR consists of some datree custom polices that are most beneficial for kubernetes users in production.
It addresses some of misconfigurations in kubernetes configuration that might sound trivial but has the potential to cause a serious impact when used in production.


eyarz:
@dhanushm7 you have some really nice use cases!

You should consider updating your rules so they will work with those resource types:

dhanushm7:
@eyarz sure, thanks for letting me know will get this updated.

dhanushm7:
@eyarz I have updated the policies and test cases that creates containers using different workloads, can you please review them again? Thanks

eyarz (approved):

PR #331, opened 2021-12-12
Hackathon 4 Custom Policy based on container and security

AdeshKhandait:
Hackathon custom policy on container and security


eyarz:
@AdeshKhandait why do you apply your policy only on resources type `Pod`?
It's also relevant for other types of resources (e.g. Deployment, DaemonSet, etc.).

I also recommend you to have one policy with multiple rules over multiple policies with one rule in each one.

AdeshKhandait:
> @AdeshKhandait why do you apply your policy only on resources type `Pod`? It's also relevant for other types of resources (e.g. Deployment, DaemonSet, etc.).
> 
> I also recommend you to have one policy with multiple rules over multiple policies with one rule in each one.

Then should i close pull request this and create a new one????

PR #333, opened 2021-12-12
feat: add custom policy example (termination-message-policy)

emjay010:
…minationMessagePolicy and terminationMessagePath is provided in k8s mainfests


eyarz:
@emjay010, you have some really nice use cases!

You should consider updating your rules, so they will work with other types of Kubernetes resources, except just `Pod`:

eyarz (approved):

PR #336, opened 2021-12-12
 docs: Cloud Native Hackathon PULL Request

NAMANIND:


eyarz:
@NAMANIND why your policy is validating that the `instances` label value is 2-4?
what is wrong with having more than 4 instances?

NAMANIND:
> @NAMANIND why your policy is validating that the `instances` label value is 2-4? what is wrong with having more than 4 instances?

While making pods if there are more instenses then it will give more load on containers. That's why I had taken instenses less than 5 such that pods have less load on it.

eyarz:

PR #337, opened 2021-12-12
Cloud Native Hackathon, Custom datree policy

chyn-1:


eyarz:
@chyn-1, good job!
The problem is that you have a rule ([CUSTOM_USER_CHECK](https://github.com/chyn-1/datree/blob/main/examples/sample-policy/policy.yaml#L29)) that will not work with resources that are creating Containers via `template` (e.g. [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment)).

You should consider updating your rule, so it will work with other types of Kubernetes resources, except just `Pod`:

eyarz:

PR #338, opened 2021-12-12
docs: adding sample custom policies for examples directory

vijaynandwani:
This PR includes policy.yaml file containing 3 rules and 8 policies. I interviewed 3 Kubernetes Administrators to come up with list of interesting policy use cases that are missing in the Datree default policy list. 

My main focus in this PR has been to create policies that cover those use cases and serve as a very solid reference example file for admins that would want to implement custom Datree policies in future.


eyarz:
@vijaynandwani, you have some really nice use cases!

You should consider updating your rules, so they will work with other types of Kubernetes resources, except just `Pod`:

eyarz:

PR #347, opened 2021-12-22
fix: add notification for invalid schema version

### LINKED ISSUE #305: Missing invalid schema version notification
**Describe the bug**

**To Reproduce**
1. Run command `datree test k8s-demo.yaml -s=1.15`
2. See error: `k8s schema validation error: could not find schema for X`

**Expected behavior**
the schema version that I passed (`1.15` instead of `1.15.0`) is invalid.
therefore, I should get some kind of indication about that.

**Screenshots**
image

 - OS: MacOs

 - Version: 0.14.49

Abhra303:
While running `datree test` , invalid argument (i.e. invalid version) to `schema-version` does not provide notification about it. Instead they get a `schema validation error`.

This PR solves that. `datree test` with invalid `schema-version` value will throw an error raising the invalid version.

Here is a screenshot -

datree-invalid-schema-version


Abhra303:
Though this approach solves the issue but I am not comfortable with this.

It is iterating through the array specified in `cmd/test/k8sValidVersions.go` and checking all the strings one by one. Which is in my opinion, time consuming.

I didn't remove the comments as I think there exist better approaches than this. Any idea how to make this efficient?

cmd/test/k8sValidVersions.go
```diff
@@ -0,0 +1,111 @@
+package test
+
+// const versions = map[string]interface{}{
```
royhadad: please remove these comments

cmd/test/main.go
```diff
@@ -146,6 +151,24 @@ func New(ctx *TestCommandContext) *cobra.Command {
 	return testCommand
 }
 
+func validateK8sVersion(k8sVersion string) error {
+	if k8sVersion == "" {
+		return nil
+	}
+//	versionPortions := strings.split(k8sVersion, ".")
```
royhadad: please remove these comments

cmd/test/k8sValidVersions.go
```diff
@@ -0,0 +1,111 @@
+package test
```
royhadad, line 1: please format both files that changed to pass the linter.
In your IDE you should have an option to format the code, if you need help with that - let me know

royhadad (changes requested):

cmd/test/k8sValidVersions.go
```diff
@@ -0,0 +1,111 @@
...
+
```
royhadad, line 2: also, please update the branch from main
Screen Shot 2021-12-22 at 14 15 17

royhadad:
> 

This pr works! which is great!
Just fix the formatting and delete the comments - and we can merge this!

Although I don't think it's that necessary, you could improve the searching algorithm in a separate PR (perhaps by using a dictionary), thank you for your contribution 😊

cmd/test/k8sValidVersions.go
Abhra303: Actually I was using Vim editor for this. Don't know, if it has any options for formatting.
I will format it in the vscode now :)

cmd/test/k8sValidVersions.go
Abhra303: okay.

Abhra303: done

royhadad (approved):

PR #355, opened 2022-01-02
fix: changelog get second latest

shmuelsa:


myishay (commented):
looking great! I think it would be better to use `tag_name`, so you can just make the changes and merge

myishay (approved):

PR #356, opened 2022-01-02
fix: show k8s schema validation error messages on each line separate

### LINKED ISSUE #350: Show all possible errors during the `Kubernetes schema validation` step
**Describe the bug**
When a file includes more than one K8s schema issues it's represented as one error (string) in the `Kubernetes schema validation` step

**To Reproduce**
1. Take the following config and run Datree against it:
```
apiVersion: v1
kind: pod
metaData:
  name: static-web
  labels:
    role: myrole
spec:
  containers:
    - name: web
      image: nginx
      ports:
        - name: web
          containerPort: 80
          protocol: TCP
```
2. See the error: `k8s schema validation error: For field (root): Additional property metaData is not allowed - For field kind: kind must be one of the following: "Pod"`

**Expected behavior**
Two separated errors:
a. `For field kind: kind must be one of the following: "Pod"`
b. `For field (root): Additional property metaData is not allowed`

**Screenshots**
image

Not relevant.

 - Version: 0.14.71

**Additional context**  
More info can be found [here](https://github.com/datreeio/datree/issues/116#issuecomment-1001624005).

amustaque97:
Fixes #350 

**Changes Proposed in PR:**
- In order to print structured error output I have added a new line in `func (e *InvalidK8sSchemaError) Error() string` method.
- Inside `ValidateResource` method replacing `-` with newline.

**Testing Instructions**
1. Create a yaml file with below content.
```yaml
apiVersion: v1
kind: pod
metaData:
  name: static-web
  labels:
    role: myrole
spec:
  containers:
    - name: web
      mage: nginx
      ports:
        - name: web
          containerPort: 80
          protocol: TCP

```
2. checkout to this branch
3.  build project
4. now run `./datree test <test-config.yaml`
5. Verify each error is coming on separate line.

**Screenshots**
Screenshot 2022-01-03 at 2 26 42 AM

cc @myishay @eyarz

### DIFF
--- a/bl/validation/k8sValidator.go
+++ b/bl/validation/k8sValidator.go
@@ -125,8 +125,15 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, error
 		// File starts with ---, the parser assumes a first empty resource
 		if res.Status == kubeconformValidator.Invalid || res.Status == kubeconformValidator.Error {
 			isValid = false
-			errorMessage := strings.ReplaceAll(res.Err.Error(), "-", "\n   ")
-			validationErrors = append(validationErrors, &InvalidK8sSchemaError{ErrorMessage: errorMessage})
+			errorMessages := strings.Split(res.Err.Error(), "-")
+
+			// errorMessages slice is not empty
+			if len(errorMessages) > 0 {
+				for _, errorMessage := range errorMessages {
+					msg := strings.Trim(errorMessage, " ")
+					validationErrors = append(validationErrors, &InvalidK8sSchemaError{ErrorMessage: msg})
+				}
+			}
 		}
 	}
 
--- a/bl/validation/k8sValidator.go
+++ b/bl/validation/k8sValidator.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/datreeio/datree/pkg/extractor"
 	kubeconformValidator "github.com/yannh/kubeconform/pkg/validator"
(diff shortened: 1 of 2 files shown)


bl/validation/k8sValidator.go
```diff
@@ -124,7 +125,8 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, error
 		// File starts with ---, the parser assumes a first empty resource
 		if res.Status == kubeconformValidator.Invalid || res.Status == kubeconformValidator.Error {
 			isValid = false
-			validationErrors = append(validationErrors, &InvalidK8sSchemaError{ErrorMessage: res.Err.Error()})
+			errorMessage := strings.ReplaceAll(res.Err.Error(), "-", "\n   ")
```
myishay: I think it would be better to split this string into different "messages" and add each message as an element to the `validationErrors` array

myishay (changes requested):
@amustaque97 Thank you for your contribution!

I gave you one comment I think should be changed.
With my suggested changes in the comment, you will probably see an outcome that prints `❌  k8s schema validation error:` several times (for each validation error). If you have an idea for printing this line only once (like you suggested in the above-mentioned image) it could be great, otherwise I think it will be ok printing it for each validation error.

amustaque97:
I have addressed review comment. Final output look like this:

Screenshot 2022-01-03 at 6 48 04 PM

myishay:
@amustaque97 LGTM!
I'll update your branch with the latest changes and merge.

myishay (approved):

PR #405, opened 2022-01-30
feat: print version message as a part of prerune hook

### LINKED ISSUE #386: Print version messages using Cobra PreRun Hook

Use Cobra Hooks before the command to check for version messages.

Resources: [https://umarcor.github.io/cobra/#prerun-and-postrun-hooks](https://umarcor.github.io/cobra/#prerun-and-postrun-hooks)

**”How to implement” Suggestion**
We print version messages in the same way on every command. For instance, you can find the logic in `datree test`  inside `evaluate` function located in `cmd/test/main.go`  (look for `LoadVersionMessages` 😬).

Eventually, adding the logic with PreRun hook will look more or less like this:

```go
&cobra.Command{
		PreRunE: func(cmd *cobra.Command, args []string) error {
			// the logic
		},
		...
		RunE: func(cmd *cobra.Command, args []string) error {...},
	}
```

**Requirements**
Golang basic level.

amustaque97:
Fixes: #386

**Description**
Run `LoadVersionMessages` as a part of PreRunE hook

**Type of changes**
Refactor/Enhancements

### DIFF
--- a/cmd/publish/main.go
+++ b/cmd/publish/main.go
@@ -12,7 +12,7 @@ import (
 )
 
 type Messager interface {
-	LoadVersionMessages(messages chan *messager.VersionMessage, cliVersion string)
+	LoadVersionMessages(cliVersion string) chan *messager.VersionMessage
 }
 
 type Printer interface {
@@ -54,14 +54,16 @@ func New(ctx *PublishCommandContext) *cobra.Command {
 			return nil
 		},
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			messages := make(chan *messager.VersionMessage, 1)
-			go ctx.Messager.LoadVersionMessages(messages, ctx.CliVersion)
-			defer func() {
-				msg, ok := <-messages
-				if ok {
-					ctx.Printer.PrintMessage(msg.MessageText+"\n", msg.MessageColor)
-				}
-			}()
+			outputFlag, _ := cmd.Flags().GetString("output")
+			if (outputFlag != "json") && (outputFlag != "yaml") && (outputFlag != "xml") {
+
+				messages := ctx.Messager.LoadVersionMessages(ctx.CliVersion)
+				defer func() {
+					for msg := range messages {
+						ctx.Printer.PrintMessage(msg.MessageText+"\n", msg.MessageColor)
+					}
+				}()
+			}
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/publish/main.go
+++ b/cmd/publish/main.go
@@ -58,11 +58,9 @@ func New(ctx *PublishCommandContext) *cobra.Command {
 			if (outputFlag != "json") && (outputFlag != "yaml") && (outputFlag != "xml") {
 
 				messages := ctx.Messager.LoadVersionMessages(ctx.CliVersion)
-				defer func() {
-					for msg := range messages {
-						ctx.Printer.PrintMessage(msg.MessageText+"\n", msg.MessageColor)
-					}
-				}()
+				for msg := range messages {
+					ctx.Printer.PrintMessage(msg.MessageText+"\n", msg.MessageColor)
+				}
 			}
 			return nil
 		},
--- a/cmd/publish/main.go
+++ b/cmd/publish/main.go
@@ -53,9 +53,7 @@ func New(ctx *PublishCommandContext) *cobra.Command {
 			}
 			return nil
 		},
-		RunE: func(cmd *cobra.Command, args []string) error {
-			cmd.SilenceUsage = true
-			cmd.SilenceErrors = true
+		PreRunE: func(cmd *cobra.Command, args []string) error {
 			messages := make(chan *messager.VersionMessage, 1)
 			go ctx.Messager.LoadVersionMessages(messages, ctx.CliVersion)
 			defer func() {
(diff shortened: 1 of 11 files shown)


dimabru:
@myishay 
can you have a look at this pr?

dimabru:
hey @amustaque97,
Thanks for your pr. Our team will review it as soon as possible.
Please note that it might take a few days

noaabarki:
Hey @amustaque97. Great! I reviewed your PR and I think there are a few things we can do to make it better :)
- Avoid fetching & printing the messages when `--output=json/XML/YAM`L:
We primarily use LoadVersionMessages to print deprecation and important messages for our users.  At the moment, if the user added the `--output=json/XML/YAML` flag they won’t be printed. As we discussed before in the issue, in the future we’d like to print them as well however, until then we want to keep the current behavior. I suggest wrapping the entire logic for fetching and printing the messages with `if (flags.Output != "json") && (flags.Output != "yaml") && (flags.Output != "xml")` inside the `PreRunE` hook.
- Fetch & Print messages in one function and call it in all commands:
It’s the same logic everywhere and it feels a little redundant. I suggest wrapping it in a `func(ctx...)`. For that, I advise changing the `LoadVersionMessages` signature into `LoadVersionMessages(cliVersion string) chan *VersionMessage` to create the messages channel and return it. (It’s also good practice in Go).
Eventually, this should be like this ->
```
PreRun: func(cmd *cobra.Command, args []string) {
			if (testCommandFlags.Output != "json") && (testCommandFlags.Output != "yaml") && (testCommandFlags.Output != "xml") {
				// this should be in a separate function
				messages := ctx.Messager.LoadVersionMessages(ctx.CliVersion)
				defer func() {
					for msg := range messages {
						ctx.Printer.PrintMessage(msg.MessageText+"\n", msg.MessageColor)
					}
				}()
			}
		}
```

amustaque97:
Thank you for the in detail review. Please note: I'm currently occupied with some other work. Might be late to address review comments.
Thank you

noaabarki:
Hey @amustaque97 , thanks for your PR 👑 We will review your PR soon, please note that it might take a day or two 🙏🏼

cmd/publish/main.go
```diff
@@ -46,17 +46,22 @@ func New(ctx *PublishCommandContext) *cobra.Command {
 			}
 			return nil
 		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			outputFlag, _ := cmd.Flags().GetString("output")
```
noaabarki, line 57: 👑

noaabarki (changes requested):
Overall looks awesome. To avoid redundant code, I suggest to write one global function that all commands can use as a util for loading the messages and printing them. WDYT?

cmd/publish/main.go
amustaque97: Hi @noaabarki, first of all Congratulations for 2022 People's Choice Award for [Kubernetes](https://opensource.com/tags/kubernetes) :tada:.

I agree with your point :100:! As per my observation refactoring will require a lot of changes and can go out of scope of this PR. Example: I was going through the `cmd/root.go` file and realised we can refactor code for all of the commands. Then we may need to update test cases as well. That's a lot of work!

My suggestion is to create a new issue and work on it.  Looking forward to hearing from you!

noaabarki (approved):

cmd/publish/main.go
noaabarki: Sounds wonderful! Do you want to take the lead on this issue?

cmd/publish/main.go
```diff
@@ -46,17 +46,22 @@ func New(ctx *PublishCommandContext) *cobra.Command {
...
+			if (outputFlag != "json") && (outputFlag != "yaml") && (outputFlag != "xml") {
+
+				messages := ctx.Messager.LoadVersionMessages(ctx.CliVersion)
+				defer func() {
```
noaabarki: Maybe I'm wrong here, but I'm not sure we need the defer.  WDYT?

cmd/test/main.go
```diff
@@ -107,6 +107,19 @@ func New(ctx *TestCommandContext) *cobra.Command {
 			}
 			return nil
 		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			outputFlag, _ := cmd.Flags().GetString("output")
+			if (outputFlag != "json") && (outputFlag != "yaml") && (outputFlag != "xml") {
+
+				messages := ctx.Messager.LoadVersionMessages(ctx.CliVersion)
+				defer func() {
```
noaabarki: Same comment about the defer.

noaabarki (changes requested):
I'm sorry for the late response. 🙏🏼 One small comment and I believe we'll be ready to go :)

cmd/publish/main.go
amustaque97: Hi @noaabarki , yeah sure I can take it up.

cmd/publish/main.go
amustaque97: No you're not wrong! It works the same without `defer` and we're not doing any work inside `PreRunE` function that require us make sure at the end of function we print messages using `defer`. I'm a noobie in golang 😅  not sure if this was written keeping some edge case in mind. 

I believe it is safe to remove `defer`. If you think same I can make code changes and submit the PR for review.

Thank You

noaabarki: Yes, let's do it.

noaabarki:
LGTM. After changes will be submitted we can merge this PR 🥳

cmd/publish/main.go
amustaque97: Made changes.

noaabarki (approved):

PR #411, opened 2022-02-08
feat: implement simple output for CI env

shmuelsa:


bl/evaluation/evaluator.go
```diff
@@ -16,12 +17,14 @@ type CLIClient interface {
 type Evaluator struct {
 	cliClient CLIClient
 	osInfo    *OSInfo
+	ciContext *ciContext.CIContext
```
noaabarki, line 20: I'm not sure we need `ciContext` in the evaluator(TBH, maybe I added the `osInfo` and I don't think it was a good idea either). 
The reason why is because we need both of them only for `CreateEvaluation` so there's no point to keep them in the evaluator context.

noaabarki: I'm not sure we need `ciContext` in the evaluator(TBH, maybe I added the `info` and I don't think it was a good idea either). The reason why is because we need both of them only for CreateEvaluation so there's no point to keep them in the evaluator context. 

I suggest creating them both in the `CreateEvaluation` function or passing them as parameters(params might help with testing). 

WDYT?

bl/evaluation/evaluator_test.go
```diff
@@ -160,6 +167,10 @@ func request_evaluation_all_valid() *evaluateTestCase {
 				PlatformVersion: "1.2.3",
 				KernelVersion:   "4.5.6",
 			},
+			ciContext: &ciContext.CIContext{
```
noaabarki: Are we testing this? Do we want to?

cmd/test/main.go
```diff
@@ -69,6 +73,12 @@ func (flags *TestCommandFlags) Validate() error {
 
 }
 
+func (flags *TestCommandFlags) Transform() {
```
noaabarki: You can set the default value when you register the flag, for example, this set "simple" a the default value: 
```
testCommand.Flags().StringP("output", "o", "simple", "Define output format")
```
wrap it in an if clause and you got yourself a deal-krokodil

pkg/ciContext/ciContext.go
```diff
@@ -0,0 +1,79 @@
...
+
+	for _, key := range os.Environ() {
+		for prefix, name := range ciMapPrefix {
+			if strings.HasPrefix(key, prefix) {
+				return name
+			}
+		}
+	}
+
+	return "none"
```
noaabarki, line 73: Simple "" is more predictable with strings IMHO 😊

noaabarki (changes requested):

bl/evaluation/evaluator_test.go
```diff
@@ -78,11 +79,15 @@ func TestCreateEvaluation(t *testing.T) {
...
+			CIEnv: "travis",
+		}
 
 		mockedCliClient.On("CreateEvaluation", mock.Anything).Return(&cliClient.CreateEvaluationResponse{EvaluationId: 1, K8sVersion: k8sVersion}, nil)
 
 		expectedCreateEvaluationResponse := &cliClient.CreateEvaluationResponse{EvaluationId: 1, K8sVersion: k8sVersion}
-		createEvaluationResponse, _ := evaluator.CreateEvaluation(cliId, cliVersion, k8sVersion, policyName)
+		createEvaluationResponse, _ := evaluator.CreateEvaluation(cliId, cliVersion, k8sVersion, policyName, ciContext)
 
 		mockedCliClient.AssertCalled(t, "CreateEvaluation", mock.Anything)
```
noaabarki: We're not testing anywhere that we actually called cli-service with the right params. Would you like to add it at least for the `ciContext`?

cmd/test/main.go
```diff
@@ -165,6 +169,10 @@ func New(ctx *TestCommandContext) *cobra.Command {
 				return err
 			}
 
+			if testCommandFlags.Output == "" && ciContext.Extract().IsCI {
```
noaabarki: Not sure we need it, because you added the parsing in line 184.

noaabarki (changes requested):

myishay (approved):

noaabarki (approved):

PR #416, opened 2022-02-14
feat: support DATREE_SCHEMA_VERSION env var

### LINKED ISSUE #414: Add env variable for schema version
Priority order:
1. user flag (-s)
2. env var
3. default version of the user account

dimabru:


cmd/test/main.go
```diff
@@ -205,6 +205,10 @@ func test(ctx *TestCommandContext, paths []string, flags TestCommandFlags) error
 		return err
 	}
 
+	if flags.K8sVersion == "" {
```
noaabarki: I wonder if this function should be responsible for the flags. I'm starting to think the flags should be outside of this function scope because: 
1.`func test(..)` already receives the flags, overriding them feels a little bit redundant and therefore difficult to understand/unexpected for the next developers. 
2.  The next developer who'd need to override a flag wouldn't know where to do it, looks like we have too many places(in addition to Shmuel PR).

I suggest having one place for binding the flags with the commands and one flag for overriding them with the local configuration. Maybe even the PreCommit hook can work for us on this one?

WDYT?

noaabarki (changes requested):

cmd/test/main.go
dimabru: Yeah, I agree.
At the moment we're using [viper](https://github.com/spf13/viper) as the configuration management tool. It does support this type of behavior. 
The problem is that making that changes and adding flag binding to viper will force a pretty big change that I think should be its own pr (this will require us to forward `*Cobra.cmd` to all `GetLocalConfiguration()` usage locations which will be a very big change.
I thought about making the change in this pr but I believe we should add a refactor task for the flag and configuration management

noaabarki: Following our conversation, we'll check moving the the local configuration override logic to the outside of `test()`. We'll do it by creating `options`(name is not mandatory) which will contain a computed options of the command flags and local configuration.

cmd/test/main.go
```diff
@@ -195,6 +211,24 @@ func New(ctx *TestCommandContext) *cobra.Command {
 	return testCommand
 }
 
+func generateTestCommandOptions(testCommandFlags *TestCommandFlags, localConfigContent *localConfig.ConfigContent) *TestCommandOptions {
```
noaabarki, line 214: I think this can me part of the TestCommandOptions struct. I'm not sure thought if we want it to be public.

noaabarki (dismissed):

cmd/test/main.go
dimabru: I think I'll keep it here for now

hadar-co (approved):

PR #418, opened 2022-02-16
fix:import output for specific CI environment

shmuelsa:


pkg/ciContext/ciContext.go
```diff
@@ -56,7 +72,7 @@ func Extract() *CIContext {
 }
 
 func isCI() string {
-	for env, name := range ciMap {
```
myishay, line 59: if environment is part of `hideEmojis` it is still a CI env

pkg/ciContext/ciContext.go
```diff
@@ -56,7 +72,7 @@ func Extract() *CIContext {
 }
 
 func isCI() string {
```
myishay: btw the function name is a bit confusing now. I would expect `isCI` to return boolean but it returns string

pkg/ciContext/ciContext.go
```diff
@@ -1,3 +1,9 @@
+/*
```
alexfedin: I would avoid comments that tied to implementation because if anything in the implementation changes the comment might not be updated and then it just becomes confusing and loses its' meaning.

pkg/ciContext/ciContext.go
```diff
@@ -73,6 +89,16 @@ func isCI() string {
 	return "none"
 }
 
+func HideEmojis() bool {
```
alexfedin: I suggest renaming to ShouldHideEmojis for better readability, HideEmojis sounds as if what it does is actually hiding the emojis

alexfedin (changes requested):

alexfedin (dismissed):

myishay (approved):

PR #419, opened 2022-02-16
feat: change test command to exit(2) when miconfigurations were found

### LINKED ISSUE #415: Add exit code/s to CLI to distinguish between different results
Currently the CLI:
* Returns 0 when a test does not find any misconfigurations.
* Returns 1 in any other case.

The CLI should:
* Return a certain error code (i.e. 1) when a test finds misconfigurations in a file.
* Return a different error code (i.e. 2) when it is used incorrectly, i.e. when an invalid argument is passed to it.

hadar-co:


dimabru (dismissed):

cmd/test/main.go
```diff
@@ -281,15 +284,15 @@ func test(ctx *TestCommandContext, paths []string, flags TestCommandFlags) error
...
-
 	if err != nil {
-		invocationFailedErr = err
-	} else if validationManager.InvalidYamlFilesCount() > 0 || validationManager.InvalidK8sFilesCount() > 0 || results.EvaluationResults.Summary.TotalFailedRules > 0 {
-		invocationFailedErr = fmt.Errorf("")
+		return err
 	}
 
-	return invocationFailedErr
+	if validationManager.InvalidYamlFilesCount() > 0 || validationManager.InvalidK8sFilesCount() > 0 || results.EvaluationResults.Summary.TotalFailedRules > 0 {
```
alexfedin: extract this to function with a clear name of what it does, it looks like there is a meaning here as to why you return the violations error but I have to read the whole thing to get it

alexfedin (changes requested):

main.go
```diff
@@ -1,13 +1,21 @@
...
 
 import (
+	"errors"
 	"os"
 
 	"github.com/datreeio/datree/cmd"
+	"github.com/datreeio/datree/cmd/test"
 )
 
+const DEFAULT_ERR_EXIT_CODE = 1
```
noaabarki, line 11: Maybe we can change the new error(ViolationsFoundError) into struct so it will contain its own exit code.

cmd/test/main.go
```diff
@@ -207,6 +210,10 @@ func validateK8sVersionFormatIfProvided(k8sVersion string) error {
 	}
 }
 
+func WereViolationsFound(validationManager *ValidationManager, results *evaluation.ResultType) bool {
```
noaabarki: I suggest to place it at the bottom and rename it to `wereViolationsFound` with lower case `w`, this way it will be like a private function

noaabarki: And also, how about validating the opposite and rename it to "invocationsPassed"? @MentalBrake @hadar-co

noaabarki (changes requested):

cmd/test/main.go
hadar-co: > And also, how about validating the opposite and rename it to "invocationsPassed"? @MentalBrake @hadar-co

@noaabarki since the validation is in sync with the error we are returning logic-wise, I think we should leave it as is, it is much clearer. I will move the function to the bottom as you suggested :)

noaabarki (approved):

alexfedin (approved):

Issue #379, opened 2022-01-17
Register test command flags with pointers 


noaabarki:
**Describe the solution you'd like**
Cleanup code and avoid unnecessary allocation by register the flags in separated function and using Cobra’s `StringVar, BoolVar...` for pointers.

**Requirements** 
Golang basic level.

**“How to implement” Suggestion**
In `cmd/test/main.go`
1. Add constructor to `TestCommandFlags`.
    
    ```go
    func NewTestCommandFlags() *TestCommandFlags {
    	return &TestCommandFlags{
    		...
    	}
    }
    ```
    
2. 2. Add  `AddFlags` method to `TestCommandFlags` that accepts cobra’s command as argument and registers flags for the command and bind the flags to `TestCommandFlags`.
    
    ```go
    type TestCommandFlags struct {
    	Output               string
    	K8sVersion           string
    	IgnoreMissingSchemas bool
    	....
    }
    
    // AddFlags registers flags for a cli
    func (flags *TestCommandFlags) AddFlags(cmd *cobra.Command) {
    	cmd.Flags().StringVarP(&flags.Output, "output", "o", "", "Define output format")
    	cmd.Flags().StringVarP(&flags.K8sVersion, "schema-version", "s", "", "Set kubernetes version to validate against. Defaults to 1.18.0")
    	cmd.Flags().StringVarP(&flags.PolicyName, "policy", "p", "", "Policy name to run against")
    	cmd.Flags().BoolVarP(&flags.OnlyK8sFiles, "only-k8s-files", "", false, "Evaluate only valid yaml files with the properties 'apiVersion' and 'kind'. Ignore everything else")
    
    	// kubeconform flags
    	cmd.Flags().StringArrayVarP(&flags.SchemaLocations, "schema-location", "", []string{}, "Override schemas location search path (can be specified multiple times)")
    	cmd.Flags().BoolVarP(&flags.IgnoreMissingSchemas, "ignore-missing-schemas", "", false, "Ignore missing schemas when executing schema validation step")
    }
    ```
    
3. Create `NewTestCommandFlags` and register using the new function you have created. 
    
    ```go
    
    func New(ctx *TestCommandContext) *cobra.Command {
    	--> testCommandFlags := NewTestCommandFlags()
    	testCommand := &cobra.Command{
    		...
    	}
    
    	--> testCommandFlags.AddFlags(testCommand)
    	return testCommand
    }
    
    ```

Kanika637:
[...]

PR #424, opened 2022-02-21
refactor: register test command flags with pointers #379

imrushi:
Code Cleanup and avoid unnecessary allocation by registering the flags in separated functions.

### DIFF
--- a/cmd/test/main.go
+++ b/cmd/test/main.go
@@ -50,6 +50,18 @@ type TestCommandFlags struct {
 	SchemaLocations      []string
 }
 
+// TestCommandFlags constructor
+func NewTestCommandFlags() *TestCommandFlags {
+	return &TestCommandFlags{
+		Output:               "",
+		K8sVersion:           "",
+		IgnoreMissingSchemas: false,
+		OnlyK8sFiles:         false,
+		PolicyName:           "",
+		SchemaLocations:      make([]string, 0),
+	}
+}
+
 //
 
 func (flags *TestCommandFlags) Validate() error {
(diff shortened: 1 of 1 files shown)


cmd/test/main.go
```diff
@@ -194,20 +207,24 @@ func New(ctx *TestCommandContext) *cobra.Command {
...
 	}
+	testCommandFlags.AddFlags(testCommand)
+	return testCommand
+}
 
+// AddFlags registers flags for a cli
+func (flags *TestCommandFlags) AddFlags(cmd *cobra.Command) {
 	if ciContext.Extract() != nil && ciContext.Extract().CIMetadata.ShouldHideEmojis {
-		testCommand.Flags().StringP("output", "o", "simple", "Define output format")
+		cmd.Flags().StringVarP(&flags.Output, "output", "o", "simple", "Define output format")
```
noaabarki, line 228: Great! I suggest moving this logic into a dedicated method of `TestCommandFlags`(maybe name it `AddFlags`?). If you have any questions about how to do it, don't hesitate to reach out 👍🏼

noaabarki (changes requested):

cmd/test/main.go
imrushi: Could you please elaborate more about the above because I have already added flag logic inside `AddFlags` Or do you want that `output` flag logic in another function and call inside `AddFlags`?

noaabarki: Oh, I'm terribly sorry, I misread it. Great job!

noaabarki (approved):

PR #429, opened 2022-02-23
refactor: move extract yamls to validator

Ben-Zaad:


pkg/yamlSchemaValidator/validator.go
```diff
@@ -14,6 +15,24 @@ func New() *YamlSchemaValidator {
 
 type Result = gojsonschema.Result
 
+func (v *YamlSchemaValidator) ExtractYamlFilesContent(schemaPath string,yamlPath string) (string, string, error) {
```
alexfedin: it's more fitting to put this in the command module as a function

YamlSchemaValidator should only do yaml schema validation, not extract yaml contents



[12:31](https://datree.slack.com/archives/D02HCFTTUP4/p1645612287530459)
YamlSchemaValidator should only do yamlSchemaValidation not extract yaml contents

alexfedin (changes requested):

alexfedin (approved):

PR #448, opened 2022-02-26
feat: add config get command

### LINKED ISSUE #440: Allow getting token with datree config command

Add an option to retrieve the token from local config. Example: 

```yaml
datree config get token
```

**Requirements**
Golang basic level.

1. Create get function for command `pkg/localConfig`. 
2. Add the `get` subcommand to the publish command(`cmd/config`).
3. Support new option with `cobra.Command` Examples and add an example for how to use this option. 
4. Add unit test (`cmd/config_test`) for running `datree config get token`, as well as the set function in `localConfig` package.

vbh:
Closes #440


cmd/config/get.go
```diff
@@ -0,0 +1,57 @@
...
+	"reflect"
+
+	"github.com/spf13/cobra"
+)
+
+func NewGetCommand(ctx *ConfigCommandContext) *cobra.Command {
+	getCommand := &cobra.Command{
+		Use:   "get",
+		Short: "Get configuration value",
+		Long:  `Get value for specific key from datree config.yaml file`,
```
shalev007: I think it'll be better to add `Defaults to $HOME/.datree/config.yaml`, same as the Set command

cmd/config/get_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	ctx := &ConfigCommandContext{
+		Messager:    messager,
+		CliVersion:  "1.2.3",
+		Printer:     printerMock,
+		LocalConfig: localConfigMock,
+	}
+
+	get(ctx, "testkey")
+	localConfigMock.AssertCalled(t, "Get", "testkey")
```
shalev007: looks great! but also maybe it's also a good idea to add a test to see that `testkey` was also printed

cmd/config/get.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+
+			msg, ok := <-messages
+			if ok {
+				ctx.Printer.PrintMessage(msg.MessageText+"\n", msg.MessageColor)
+			}
+		},
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("requires exactly 1 arguments")
```
shalev007:
```suggestion
				return errors.New("requires exactly 1 argument")
```

cmd/config/get.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+
+			validKeys := make(map[string]bool)
+			validKeys["token"] = true
+
+			if val, ok := validKeys[args[0]]; !ok || !val {
```
shalev007: I think since this validation also appears on the `set` command, can we outsource this validation as a function and use it in both commands?

@dimabru, your thoughts?

dimabru: I agree. Adding 2 `validKeys` objects that we need to maintain is bad practice. I think its best to keep the `validKeys` logic in one place to avoid potential bugs should we later decide to add/remove keys in config.yaml.
@vbh what do you think?

dimabru (changes requested):

cmd/config/set.go
```diff
@@ -31,20 +31,23 @@ func NewSetCommand(ctx *ConfigCommandContext) *cobra.Command {
...
-
-			return nil
+			return validateKey(args[0])
 		},
 	}
 
 	return setCommand
 }
 
+func validateKey(key string) error {
```
shalev007: Everything looks great!
But before we approve, I don't feel like this function's place is on the `set.go` file since it's being consumed on both `set.go` and `get.go`, Could you move it to a more "global" place, like main or a validator file?

mashtuba:
Hi @vbh, thanks for submitting the PR! 🙌
We’re currently working on moving our evaluation process to the CLI, therefore we’re in a code-freeze until we merge these changes.
Your PR will be reviewed and merged once we’re out from code freeze, by the end of next week 🥳

adifayer:
Hi @vbh, thanks for submitting the PR! 🙌
We’re currently working on moving our evaluation process to the CLI, therefore we’re in a code-freeze until we merge these changes.
Your PR will be reviewed and merged once we’re out from code freeze, by the end of next week 🥳

dimabru (dismissed):
Hi @vbh,
due to our latest big local evaluation changes there are some conflicts in your pr. please fix them and we should be able to give it one last review before we merge

Ben-Zaad:
Hi @vbh, I'm Ben from Datree :-)
We noticed that it's been a while since your last commit, please feel free to contact me if you have any questions about the requested changes

Ben-Zaad (approved):

shalev007 (approved):

PR #452, opened 2022-02-27
perf: read files in parallel

### LINKED ISSUE #447: Read files in parallel

Currently, the way test command works is that datree iterate every file in the the given path with  for loop inside a single `gouruitine` . This is not done in parallel since the using the for loop is blocking. Instead, we want to read all files in parallel. 

**Requirements**
Golang intermediate level.

1. Convert paths array into the channel. the path channel should be buffered, use 100 as the buffer size. Adjust `filesConfigurationsChan` and `invalidFilesChan` to the same buffer size. 
2. Initiate a waiting group with the `concurrency` parameter(see `ExtractFilesConfigurations` function) and use it to wait until all paths have been read.
    1. Loop `i=0; i< concurrency` and create `goroutines`. Inside each goroutine iterate the paths channel and call `ExtractConfigurationsFromYamlFile`.
    2. Close worker when the channel is empty and closed.

vbh:
Closes #447

### DIFF
--- a/bl/files/files.go
+++ b/bl/files/files.go
@@ -3,6 +3,7 @@ package files
 import (
 	"bytes"
 	"os"
+	"sync"
 
 	"gopkg.in/yaml.v3"
 
@@ -25,6 +26,14 @@ func New() *FilesExtractor {
 func (f *FilesExtractor) ExtractFilesConfigurations(paths []string, concurrency int) (chan *extractor.FileConfigurations, chan *extractor.InvalidFile) {
 	filesConfigurationsChan := make(chan *extractor.FileConfigurations, concurrency)
 	invalidFilesChan := make(chan *extractor.InvalidFile, concurrency)
+	pathsChan := make(chan string, concurrency)
+
+	go func() {
+		defer close(pathsChan)
+		for _, path := range paths {
+			pathsChan <- path
+		}
+	}()
 
 	go func() {
 		defer func() {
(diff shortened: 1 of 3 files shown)


shalev007:
Hey, @vbh Thank you for your contribution!
We're starting to review your PR now,
I updated your branch with the latest changes

shalev007:
@noaabarki I remember you did something similar, would you like to give your thought on this?

bl/files/files.go
```diff
@@ -13,24 +15,39 @@ type UnknownStruct map[string]interface{}
 func ExtractFilesConfigurations(paths []string, concurrency int) (chan *extractor.FileConfigurations, chan *extractor.InvalidFile) {
 	filesConfigurationsChan := make(chan *extractor.FileConfigurations, concurrency)
 	invalidFilesChan := make(chan *extractor.InvalidFile, concurrency)
+	pathsChan := make(chan string, concurrency)
+
+	go func() {
```
noaabarki: For clarity, I suggest to wrap this in a private method maybe `parsePathsArrayToChan()`

bl/files/files.go
```diff
@@ -13,24 +15,39 @@ type UnknownStruct map[string]interface{}
...
+		defer close(pathsChan)
+		for _, path := range paths {
+			pathsChan <- path
+		}
+	}()
 
 	go func() {
 		defer func() {
```
noaabarki: No need to use defer, we can close both channels after `wg.Wait()`

bl/files/files.go
```diff
@@ -13,24 +15,39 @@ type UnknownStruct map[string]interface{}
...
 
-		for _, path := range paths {
-
-			configurations, absolutePath, invalidYamlFile := extractor.ExtractConfigurationsFromYamlFile(path)
+		var wg sync.WaitGroup
+		wg.Add(concurrency)
+		for i := 0; i < concurrency; i++ {
+			go func() {
+				defer wg.Done()
+				for path := range pathsChan {
```
noaabarki: How about using using a for loop with `path, ok := <- pathChan` to make sure we read form pathsChan until it's both empty and closed?

noaabarki (dismissed):
Great work @vbh! Great to have you on board with us! 🏆

mashtuba:
Hi @vbh, thanks for submitting the PR! 🙌
We’re currently working on moving our evaluation process to the CLI, therefore we’re in a code-freeze until we merge these changes.
Your PR will be reviewed and merged once we’re out from code freeze, by the end of next week 🥳

adifayer:
Hi @vbh, thanks for submitting the PR! 🙌
We’re currently working on moving our evaluation process to the CLI, therefore we’re in a code-freeze until we merge these changes.
Your PR will be reviewed and merged once we’re out from code freeze, by the end of next week 🥳

shalev007:
Hi @vbh,
Just an update,
were out of the code freeze and I'll take a look at the PR as soon as I can, so we can push this further

shalev007:
@vbh 
Sorry for the late response,
We had to validate this works for edge cases as well and found an issue,
for example:
- if we read more files than the `concurrency` variable the test is stuck 

possible fix: on `cmd/test/main.go` on line 414 please use the function as a go routine ie: `go validationManager.AggregateInvalidYamlFiles(invalidYamlFilesChan)`

Could you please add that and check edge cases such as a big amount of files? 🙏 
We're also available on our community [slack channel](https://datreecommunity.slack.com/archives/C02L0QB3YTC) for any discussion

myishay:
@vbh ping 😊
any news on this one?

shalev007:
Hey @vbh, I was testing your changes and it appears it still doesn't work when there are more files than the concurrent files
for example:
1. change the concurrency to 2 (`concurrency := 2`)
2. compile (`make build-staging`)
3. and run `./datree ./internal/fixtures/kube/empty.yaml ./internal/fixtures/kube/empty.yaml ./internal/fixtures/kube/empty.yaml`

Expected:
files will be evaluated with a yaml error

Outcome:
it never finished evaluating, the loader doesn't end ever

shalev007 (approved):

PR #453, opened 2022-02-27
feat: policy package & validator tests

dimabru:


hadar-co (dismissed):

pkg/fileReader/reader.go
```diff
@@ -80,3 +80,27 @@ func (fr *FileReader) ReadFileContent(filepath string) (string, error) {
 
 	return string(dat), nil
 }
+
+func (fr *FileReader) ReadDir(dirpath string) ([]string, error) {
+	var files []string
+
+	files, err := fr.glob(dirpath + "/**/*")
```
noaabarki, line 87: We might to validate `isDir(dirpath)`

pkg/fileReader/reader.go
```diff
@@ -80,3 +80,27 @@ func (fr *FileReader) ReadFileContent(filepath string) (string, error) {
 
 	return string(dat), nil
 }
+
+func (fr *FileReader) ReadDir(dirpath string) ([]string, error) {
+	var files []string
```
noaabarki: We might need to validate `isDir(dirpath)`

pkg/yamlSchemaValidator/validator_test.go
```diff
@@ -0,0 +1,112 @@
...
+	fails  []*FileWithName
+	passes []*FileWithName
+}
+
+type FileWithName struct {
+	name    string
+	content string
+}
+
+func TestDefaultRulesValidation(t *testing.T) {
```
noaabarki: The common convention for test function is TestXxx where Xxx is the name of the tested function/method. I suggest renaming this test because it might be confusing for the next developer because there isn't any `DefaultRulesValidation` function. IMHO, the name should indicate `Validate` and testing the default rules is one of the test cases.

pkg/yamlSchemaValidator/validator_test.go
```diff
@@ -0,0 +1,112 @@
...
+		res, err := validator.Validate(string(schemaBytes), file.content)
+		if err != nil {
+			panic(err)
+		}
+
+		if len(res.Errors()) > 0 && expectPass {
+			t.Errorf("Expected validation for rule with id %d to pass, but it failed for file %s\n", ruleId, file.name)
+		}
+		if len(res.Errors()) == 0 && !expectPass {
+			t.Errorf("Expected validation for rule with id %d to fail, but it passed for file %s\n", ruleId, file.name)
```
noaabarki, line 62: This error is thrown from `gojsonschema`? If so, this is a bit coupling and might be unexpected in the future. I suggest creating a specific error and returning it. This way we can actually use the same error and it will make more sense why we expected this error.

pkg/policy/policy.go
```diff
@@ -0,0 +1,44 @@
...
+	defaultRulesYaml, err := fileReader.ReadFileContent(defaultRulesYamlPath)
+	if err != nil {
+		return nil, err
+	}
+
+	defaultRulesDefinitions, err := yamlToStruct(defaultRulesYaml)
+	return defaultRulesDefinitions, err
+}
+
+func yamlToStruct(content string) (*DefaultRulesDefinitions, error) {
```
noaabarki, line 37: In my dreams, we use generics and move this to utils 😈

noaabarki (changes requested):
This PR was a delight to review! @dimabru your'e a champ! 🏆

alexfedin (dismissed):

pkg/policy/policy.go
```diff
@@ -0,0 +1,44 @@
...
+	Rules      []DefaultRuleDefinition `yaml:"rules"`
+}
+
+type DefaultRuleDefinition struct {
+	ID               int                    `yaml:"id"`
+	Name             string                 `yaml:"name"`
+	UniqueName       string                 `yaml:"uniqueName"`
+	EnabledByDefault bool                   `yaml:"enabledByDefault"`
+	DocumentationUrl string                 `yaml:"documentationUrl"`
+	MessageOnFailue  string                 `yaml:"messageOnFailue"`
```
TzlilSwimmer123: @dimabru `MessageOnFailue` is containing the typo :)

pkg/fileReader/reader.go
dimabru: done

pkg/yamlSchemaValidator/validator_test.go
dimabru: renamed to `TestValidate`

pkg/yamlSchemaValidator/validator_test.go
dimabru: Generally I agree but in this case `Errors()` is a very generic method name and we can easily implement the same logic even if we decide to change the package we use.
I see no reason adding our own errors implementation when the existing one is readable.
Also, the test is not using the data inside the `Errors()` array but only checking its length

pkg/policy/policy.go
dimabru: we can open source this as an issue :)

pkg/policy/policy.go
dimabru: thanks :)

noaabarki (approved):

PR #463, opened 2022-03-03
feat: update regex logic

eyarz:
https://regex101.com/r/NeqtPy/1


eyarz:
@dimabru verified with `datree schema-validator`

dimabru (approved):

PR #464, opened 2022-03-03
feat: report unexpected cli errors

royhadad:


bl/errorReporter/errorReporter.go
```diff
@@ -0,0 +1,70 @@
...
+func NewErrorReporter(client CliClient, localConfig LocalConfig) *ErrorReporter {
+	return &ErrorReporter{
+		client: client,
+		config: localConfig,
+	}
+}
+
+func (reporter *ErrorReporter) ReportCliError(panicErr interface{}) {
+	status, err := reporter.client.ReportCliError(cliClient.ReportCliErrorRequest{
+		CliId:        reporter.getCliId(),
```
dimabru: I think this needs to be `token` instead of cliId because we are already in the process of changing naming conventions in all backend requests

bl/errorReporter/errorReporter.go
```diff
@@ -0,0 +1,70 @@
...
+		StackTrace:   string(debug.Stack()),
+	})
+	if err != nil {
+		fmt.Println("failed to report unexpected panic:\nstatus code:", status, "\nerror:", err.Error())
+	} else {
+		fmt.Println("panic reported")
+	}
+}
+
+func (reporter *ErrorReporter) getCliId() string {
```
dimabru: use defer and recover if `GetLocalConfiguration` panics

dimabru (approved):

PR #472, opened 2022-03-08
feat: add create token and client id set on local config

shmuelsa:


pkg/localConfig/localConfig.go
```diff
@@ -71,6 +85,21 @@ func (lc *LocalConfig) Set(key string, value string) error {
 	return nil
 }
 
+func InitLocalConfigFile() error {
+	configHome, configName, configType, err := setViperConfig()
```
dimabru, line 100: make sure to keep the comment with the references to the issue and pr in viper

shmuelsa: added.

pkg/localConfig/localConfig.go
```diff
@@ -5,69 +5,120 @@ import (
...
+	ClientId      string
 	SchemaVersion string
 }
 
-type LocalConfig struct {
+type TokenClient interface {
+	CreateToken() (*cliClient.CreateTokenResponse, error)
+}
+
+type LocalConfigClient struct {
```
myishay, line 23: I think it should be called `LocalConfig` and not `LocalConfigClient`. this struct contains a client

myishay (dismissed):

myishay (dismissed):

myishay (dismissed):

shalev007 (approved):

PR #473, opened 2022-03-08
feat: local evaluation full implementation

TzlilSwimmer123:


bl/evaluation/printer.go
```diff
@@ -287,45 +287,45 @@ func parseEvaluationResultsToSummary(results *EvaluationResults, evaluationSumma
...
+//	// Iterate over mapper and create XML tokens for all entries
+//	for filePath, encapsulatedRule := range mapper {
+//		keys := make([]int, 0)
+//		for i := range encapsulatedRule {
+//			keys = append(keys, i)
+//		}
+//		sort.Ints(keys)
+//
+//		for _, key := range keys { // Since rule is encapsulated by its id (int), we don't add is a tag
+//			rule := encapsulatedRule[key]
```
TzlilSwimmer123: Not sure if we need this code - ATM it's unused. It can work with local evaluation.

bl/evaluation/printer.go
```diff
@@ -163,18 +163,18 @@ func parseToPrinterWarnings(results *EvaluationResults, invalidYamlFiles []*extr
...
 
-			rulesIds := []int{}
-			for ruleId := range rules {
-				rulesIds = append(rulesIds, ruleId)
+			rulesUniqueNames := []string{}
+			for rulesUniqueName := range rules {
+				rulesUniqueNames = append(rulesUniqueNames, rulesUniqueName)
 			}
-			sort.Ints(rulesIds)
+			//sort.Ints(rulesIds)
```
TzlilSwimmer123: Should add sort by rule name

bl/evaluation/evaluator.go
```diff
@@ -1,16 +1,21 @@
...
 	"github.com/datreeio/datree/pkg/cliClient"
 	"github.com/datreeio/datree/pkg/extractor"
+	"github.com/datreeio/datree/pkg/yamlSchemaValidator"
 )
 
 type CLIClient interface {
 	RequestEvaluation(request *cliClient.EvaluationRequest) (*cliClient.EvaluationResponse, error)
 	CreateEvaluation(request *cliClient.CreateEvaluationRequest) (*cliClient.CreateEvaluationResponse, error)
 	SendFailedYamlValidation(request *cliClient.UpdateEvaluationValidationRequest) error
 	SendFailedK8sValidation(request *cliClient.UpdateEvaluationValidationRequest) error
```
dimabru: delete all of the requests above

bl/evaluation/evaluator.go
```diff
@@ -42,10 +48,16 @@ type EvaluationResults struct {
...
-func (e *Evaluator) CreateEvaluation(cliId string, cliVersion string, k8sVersion string, policyName string, ciContext *ciContext.CIContext) (*cliClient.CreateEvaluationResponse, error) {
-	createEvaluationResponse, err := e.cliClient.CreateEvaluation(&cliClient.CreateEvaluationRequest{
-		K8sVersion: &k8sVersion,
-		CliId:      cliId,
+type FormattedResults struct {
+	EvaluationResults               *EvaluationResults
+	NonInteractiveEvaluationResults *NonInteractiveEvaluationResults
+}
+
+func (e *Evaluator) SendLocalEvaluationResult(cliId string, cliVersion string, k8sVersion string, policyName string, ciContext *ciContext.CIContext, rulesData []cliClient.RuleData, filesData []cliClient.FileData, failedYamlFiles []string, failedK8sFiles []string, policyCheckResults map[string]map[string]cliClient.FailedRule) (*cliClient.SendEvaluationResultsResponse, error) {
```
dimabru: all of these parameters can be contained in a single struct. its very hard to keep track of everything

bl/evaluation/evaluator.go
```diff
@@ -54,75 +66,104 @@ func (e *Evaluator) CreateEvaluation(cliId string, cliVersion string, k8sVersion
...
-		invalidFilesPaths = append(invalidFilesPaths, &file.Path)
-	}
-	err := e.cliClient.SendFailedYamlValidation(&cliClient.UpdateEvaluationValidationRequest{
-		EvaluationId:   evaluationId,
-		InvalidFiles:   invalidFilesPaths,
-		StopEvaluation: stopEvaluation,
-	})
-	return err
-}
+func (e *Evaluator) Evaluate(filesConfigurations []*extractor.FileConfigurations, isInteractiveMode bool, policyName string, policy policy_factory.Policy) (FormattedResults, []cliClient.RuleData, []cliClient.FileData, map[string]map[string]cliClient.FailedRule, int, error) {
```
dimabru: same - all parameters can go into a struct

bl/evaluation/evaluator.go
```diff
@@ -54,75 +66,104 @@ func (e *Evaluator) CreateEvaluation(cliId string, cliVersion string, k8sVersion
...
+
+		for _, configuration := range filesConfiguration.Configurations {
+			for _, rulesSchema := range policy.Rules {
+				rulesData = append(rulesData, cliClient.RuleData{Identifier: rulesSchema.RuleIdentifier, Name: rulesSchema.RuleName})
+
+				kind := configuration["kind"].(string)
+				metadata := configuration["metadata"]
+				name := metadata.(map[string]interface{})["name"].(string)
+
+				configurationJson, _ := json.Marshal(configuration)
```
dimabru: we will need to deal with the error objects inside this block of code. they should not happen and if they do - its a good place to report them

bl/evaluation/evaluator.go
```diff
@@ -54,75 +66,104 @@ func (e *Evaluator) CreateEvaluation(cliId string, cliVersion string, k8sVersion
...
+	formattedResults.EvaluationResults = e.formatEvaluationResults(failedDict, len(filesConfigurations))
+
 	if !isInteractiveMode {
-		resultType.NonInteractiveEvaluationResults = e.formatNonInteractiveEvaluationResults(resultType.EvaluationResults, res.Results, evaluationResponse.PolicyName, evaluationResponse.RulesCount)
+		formattedResults.NonInteractiveEvaluationResults = e.formatNonInteractiveEvaluationResults(formattedResults.EvaluationResults, failedDict, policyName, rulesCount)
 	}
-	return resultType, nil
+
+	if rulesData == nil {
+		rulesData = []cliClient.RuleData{}
```
dimabru: you can have the same logic in the init of `rulesData` and avoid this `if`

bl/evaluation/printer.go
dimabru: make sure its not used, then delete it

bl/policy/policy_factory.go
```diff
@@ -0,0 +1,87 @@
...
+		return Policy{}, err
+	}
+
+	var rules []RuleSchema
+
+	if policies != nil {
+		var chosenPolicy *cliClient.Policy
+		getDefaultPolicy := false
+
+		if policyName == "" || policyName == "default" {
```
dimabru: policyName is set by the user. the name `default` should not be special. should be only `if policyName == "" {`

bl/policy/policy_factory.go
```diff
@@ -0,0 +1,87 @@
...
+			return Policy{}, err
+		}
+
+		policyName = chosenPolicy.Name
+
+		if chosenPolicy.Rules == nil {
+			return Policy{policyName, []RuleSchema{}}, nil
+		}
+
+		for _, rule := range chosenPolicy.Rules {
```
dimabru: all that loop can go into a separate `populateRules` function to simplify the code a bit

bl/policy/policy_factory.go
```diff
@@ -0,0 +1,87 @@
...
+			var isCustomRule bool
+
+			for _, customRule := range policies.CustomRules {
+				if rule.Identifier == customRule.Identifier {
+					rules = append(rules, RuleSchema{rule.Identifier, customRule.Name, customRule.Schema, customRule.DefaultMessageOnFailure})
+					isCustomRule = true
+				}
+			}
+
+			if !isCustomRule {
```
dimabru: if a rule is not default nor custom we should return an error. that behavior should not happen

cmd/test/main.go
```diff
@@ -198,20 +204,29 @@ func New(ctx *TestCommandContext) *cobra.Command {
...
+			localConfigContent, err := ctx.LocalConfig.GetLocalConfiguration()
 			if err != nil {
 				return err
 			}
 
-			localConfigContent, err := ctx.LocalConfig.GetLocalConfiguration()
+			prerunDataForEvaluation, statusCode, err := ctx.CliClient.RequestPrerunDataForEvaluation(localConfigContent.CliId)
+
+			if err != nil {
+				if statusCode != 404 {
```
dimabru: I think we need to add a comment here to explain this behavior. if someone is reading this it will be unclear what's the purpose of this exception

cmd/test/main.go
```diff
@@ -303,7 +322,14 @@ func test(ctx *TestCommandContext, paths []string, options *TestCommandOptions)
 		ctx.Printer.SetTheme(printer.CreateSimpleTheme())
 	}
 
-	validationManager, createEvaluationResponse, results, err := evaluate(ctx, filesPaths, options)
+	policy, err := policy_factory.CreatePolicy(policies, options.PolicyName)
+	if err != nil {
+		return err
+	}
+
+	options.PolicyName = policy.Name
```
dimabru: `options` population should happen outside of `test` command. that way you will only forward to `test` the required policy

cmd/test/main.go
```diff
@@ -352,7 +378,7 @@ func test(ctx *TestCommandContext, paths []string, options *TestCommandOptions)
 	return nil
 }
 
-func evaluate(ctx *TestCommandContext, filesPaths []string, options *TestCommandOptions) (*ValidationManager, *cliClient.CreateEvaluationResponse, evaluation.ResultType, error) {
+func evaluate(ctx *TestCommandContext, filesPaths []string, options *TestCommandOptions, policy policy_factory.Policy) (*ValidationManager, int, evaluation.FormattedResults, string, error) {
```
dimabru: the functions' return values are unclear. we will need to be more specific about what it returns. adding `int` or `string` alongside other variables is making the code less readable. maybe add them to a struct?

pkg/cliClient/evaluation.go
```diff
@@ -106,6 +111,98 @@ func (c *CliClient) RequestEvaluation(request *EvaluationRequest) (*EvaluationRe
...
+	CustomRules []*CustomRule `json:"customRules"`
+	Policies    []*Policy     `json:"policies"`
+}
+
+type PrerunDataForEvaluationResponse struct {
+	PoliciesJson      *PrerunPoliciesForEvaluation `json:"policiesJson"`
+	DefaultK8sVersion string                       `json:"defaultK8sVersion"`
+}
+
+func (c *CliClient) RequestPrerunDataForEvaluation(tokenId string) (*PrerunDataForEvaluationResponse, int, error) {
```
dimabru: delete all other functions in this file. they are not used anymore :)

pkg/cliClient/evaluation.go
```diff
@@ -106,6 +111,98 @@ func (c *CliClient) RequestEvaluation(request *EvaluationRequest) (*EvaluationRe
...
+}
+
+type FailedRule struct {
+	Name             string          `json:"ruleName"`
+	MessageOnFailure string          `json:"messageOnFailure"`
+	Configurations   []Configuration `json:"configurations"`
+}
+
+type LocalEvaluationResultRequest struct {
+	ClientId           string                           `json:"cliId"`
```
dimabru: `json: clientId`

dimabru (changes requested):

bl/evaluation/evaluator.go
TzlilSwimmer123: goland does not like `rulesData := []cliClient.RuleData{}`. What do you suggest?

bl/evaluation/evaluator.go
TzlilSwimmer123: in this specific block of code the panic should work for the metadata and name but for the Marshel we can report...if I understand you correctly

pkg/cliClient/evaluation.go
TzlilSwimmer123: good catch

bl/evaluation/evaluator.go
```diff
@@ -42,87 +44,155 @@ type EvaluationResults struct {
...
+	PolicyName         string
+	CiContext          *ciContext.CIContext
+	RulesData          []cliClient.RuleData
+	FilesData          []cliClient.FileData
+	FailedYamlFiles    []string
+	FailedK8sFiles     []string
+	PolicyCheckResults map[string]map[string]cliClient.FailedRule
+}
+
+func (e *Evaluator) SendLocalEvaluationResult(localEvaluationRequestData LocalEvaluationRequestData) (*cliClient.SendEvaluationResultsResponse, error) {
```
alexfedin: looks like a redundant wrapper to the cliClient, let's just use the cliClient to send the results

bl/evaluation/evaluator.go
```diff
@@ -42,87 +44,155 @@ type EvaluationResults struct {
...
+type PolicyCheckResultData struct {
+	FormattedResults FormattedResults
+	RulesData        []cliClient.RuleData
+	FilesData        []cliClient.FileData
+	RawResults       map[string]map[string]cliClient.FailedRule
+	RulesCount       int
 }
 
-func (e *Evaluator) Evaluate(filesConfigurations []*extractor.FileConfigurations, evaluationResponse *cliClient.CreateEvaluationResponse, isInteractiveMode bool) (ResultType, error) {
+func (e *Evaluator) Evaluate(dataForEvaluation DataForPolicyCheck) (PolicyCheckResultData, error) {
```
alexfedin: using For or From in types and function names is tightly coupling them to the implementation of the usage, use a name that describes the the entity itself without relying on its usage

bl/evaluation/evaluator.go
```diff
@@ -42,87 +44,155 @@ type EvaluationResults struct {
...
-	res, err := e.cliClient.RequestEvaluation(&cliClient.EvaluationRequest{
-		EvaluationId: evaluationResponse.EvaluationId,
-		Files:        filesConfigurations,
-	})
-	if err != nil {
-		return ResultType{}, err
+	rulesCount := len(dataForEvaluation.Policy.Rules)
+
+	// map of files paths to map of rules to failed rule data
+	failedDict := make(map[string]map[string]cliClient.FailedRule)
```
alexfedin: I suggest renaming it to failedRulesByFiles, the comment doesn't really makes it easier to understand

bl/evaluation/evaluator.go
```diff
@@ -42,87 +44,155 @@ type EvaluationResults struct {
...
+	emptyPolicyCheckResult := PolicyCheckResultData{FormattedResults{}, []cliClient.RuleData{}, []cliClient.FileData{}, nil, 0}
+
+	for _, filesConfiguration := range dataForEvaluation.FilesConfigurations {
+		filesData = append(filesData, cliClient.FileData{FilePath: filesConfiguration.FileName, ConfigurationsCount: len(filesConfiguration.Configurations)})
+
+		for _, configuration := range filesConfiguration.Configurations {
+			for _, rulesSchema := range dataForEvaluation.Policy.Rules {
+				rulesData = append(rulesData, cliClient.RuleData{Identifier: rulesSchema.RuleIdentifier, Name: rulesSchema.RuleName})
+
+				kind := configuration["kind"].(string)
```
alexfedin: this can be extracted into a single function that receives a configuration and extracts the data you need

bl/evaluation/evaluator.go
```diff
@@ -42,87 +44,155 @@ type EvaluationResults struct {
...
+				metadata := configuration["metadata"]
+				name := metadata.(map[string]interface{})["name"].(string)
+
+				configurationJson, err := json.Marshal(configuration)
+				if err != nil {
+					return emptyPolicyCheckResult, err
+				}
+
+				yamlSchemaValidatorInst := yamlSchemaValidator.New()
```
alexfedin: do we really need to create yamlSxhemaValidator instance per schema? I think we can create it in the validator constructor and always use the same instance.

side note, no need to add Inst to the name, it's obvious that it's the instance so it doesn't add new information or context

bl/evaluation/evaluator.go
```diff
@@ -42,87 +44,155 @@ type EvaluationResults struct {
...
+				result, err := yamlSchemaValidatorInst.Validate(string(ruleSchemaJson), string(configurationJson))
+
+				if err != nil {
+					return emptyPolicyCheckResult, err
+				}
+
+				if len(result.Errors()) > 0 {
+					configurationData := cliClient.Configuration{Name: name, Kind: kind, Occurrences: len(result.Errors())}
+
+					if fileData, ok := failedDict[filesConfiguration.FileName]; ok {
```
alexfedin: I think we can extract the creation of the failures object into its own function, as of now the whole evaluate function is too long to easily follow what it does

bl/evaluation/evaluator_test.go
```diff
@@ -15,14 +15,14 @@ type mockCliClient struct {
 	mock.Mock
 }
 
-func (m *mockCliClient) CreateEvaluation(createEvaluationRequest *cliClient.CreateEvaluationRequest) (*cliClient.CreateEvaluationResponse, error) {
-	args := m.Called(createEvaluationRequest)
-	return args.Get(0).(*cliClient.CreateEvaluationResponse), args.Error(1)
+func (m *mockCliClient) RequestPrerunDataForEvaluation(token string) (*cliClient.PrerunDataForEvaluationResponse, error) {
```
alexfedin: Same regarding for in function names, this should be just RequestPrerunData

bl/evaluation/evaluator_test.go
```diff
@@ -15,14 +15,14 @@ type mockCliClient struct {
...
+	args := m.Called(token)
+	return args.Get(0).(*cliClient.PrerunDataForEvaluationResponse), args.Error(1)
 }
 
-func (m *mockCliClient) RequestEvaluation(evaluationRequest *cliClient.EvaluationRequest) (*cliClient.EvaluationResponse, error) {
-	args := m.Called(evaluationRequest)
-	return args.Get(0).(*cliClient.EvaluationResponse), args.Error(1)
+func (m *mockCliClient) SendLocalEvaluationResult(localEvaluationResultRequest *cliClient.LocalEvaluationResultRequest) (*cliClient.SendEvaluationResultsResponse, error) {
```
alexfedin: The rules evaluation will always be local, making this information redundant, I suggest renaming to SendEvaluationResult

bl/evaluation/evaluator_test.go
```diff
@@ -119,21 +142,20 @@ func TestEvaluate(t *testing.T) {
 
 			mockedCliClient.On("RequestEvaluation", mock.Anything).Return(tt.mock.cliClient.requestEvaluation.response, tt.mock.cliClient.requestEvaluation.err)
 
-			evaluator := &Evaluator{
-				cliClient: mockedCliClient,
-				osInfo:    tt.args.osInfo,
-			}
+			//evaluator := &Evaluator{
```
alexfedin: Delete unused comments, we have git if we will want to recover it someday

bl/evaluation/printer.go
```diff
@@ -163,18 +163,18 @@ func parseToPrinterWarnings(results *EvaluationResults, invalidYamlFiles []*extr
...
 
-			for _, ruleId := range rulesIds {
-				rule := rules[ruleId]
+			for _, rulesUniqueName := range rulesUniqueNames {
```
alexfedin: *ruleUniqueName or just uniqueName

cmd/test/main.go
```diff
@@ -190,7 +197,18 @@ func New(ctx *TestCommandContext) *cobra.Command {
 				return err
 			}
 
-			testCommandOptions := GenerateTestCommandOptions(testCommandFlags, localConfigContent)
+			prerunDataForEvaluation, statusCode, err := ctx.CliClient.RequestPrerunDataForEvaluation(localConfigContent.CliId)
+
+			// getting prerun data can return 404 if user has a valid token but he didn't sign up yet - suppose to be ok
```
alexfedin: you can add this data inside the client, so it will return the results according to the logic written in the comments, that way it will be integrated in the code and not be hidden in the comments, also if we will want to use it in another place we will have to rewrite the same logic in the new place

alexfedin (changes requested):

bl/evaluation/evaluator.go
TzlilSwimmer123: That was the convention until now, I thought to keep with it...

bl/evaluation/evaluator.go
TzlilSwimmer123: I agree it's not the right place but what is the validator constructor?

bl/evaluation/evaluator_test.go
TzlilSwimmer123: I fixed the tests - now it's not commented in

dimabru (dismissed):

alexfedin (approved):

PR #487, opened 2022-03-14
fix: containers_missing_memory_request

Ben-Zaad:


royhadad:
fixed in: https://github.com/datreeio/datree/pull/490

PR #491, opened 2022-03-14
fix: set default k8s version if not received from config or from server

alexfedin:


cmd/test/main.go
```diff
@@ -244,6 +244,10 @@ func GenerateTestCommandData(testCommandFlags *TestCommandFlags, localConfigCont
 		k8sVersion = evaluationPrerunDataResp.DefaultK8sVersion
 	}
 
+	if k8sVersion == "" {
```
TzlilSwimmer123, line 247: Maybe we can refactor it to be in the first `if k8sVersion == "" `? now we repeat on it twice

royhadad (approved):

cmd/test/main.go
alexfedin: it doesn't make sense, the priority goes env var > server response > default

PR #498, opened 2022-03-15
feat: add policiesSchema.yaml

Ben-Zaad:


alexfedin (dismissed):

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
...
+            "type": "boolean"
+          },
+          "message": {
+            "type": "string"
+          },
+          "rules": {
+            "items": {
+              "type": "object",
+              "properties": {
+                "identifier": {
```
royhadad, line 32: missing property "messageOnFailure"

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
...
+                  "type": "string",
+                  "pattern": "^([A-Z0-9])+(_?[A-Z0-9])*$"
```
royhadad: to the best of my knowledge, we don't have this regex limitation in our backend validator

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
...
+              ]
+            }
+          }
+        }
+      }
+    },
+    "customRules": {
+      "type": "array",
+      "items": {
+        "identifier": "string",
```
royhadad: I think you meant to write:
```
type: object
properties: {
    identifier: "string"
}
```

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
...
+        "pattern": "^([A-Z0-9])+(_?[A-Z0-9])*$",
```
royhadad: incorrect regex, this is what we have in backend:
Screen Shot 2022-03-16 at 14 48 04

pkg/policy/policiesSchema.json
royhadad: we need to check that identifier is not in default rules, this is from backend:
Screen Shot 2022-03-16 at 14 48 42

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
+{
+  "type": "object",
+  "properties": {
+    "apiVersion": {
+      "type": "string",
+      "enum": [
+        "v1"
+      ]
+    },
+    "message": {
```
royhadad: what is this key "message"?

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
...
+        "identifier": "string",
+        "pattern": "^([A-Z0-9])+(_?[A-Z0-9])*$",
+        "defaultMessageOnFailure": {
+          "type":"string"
+        },
+        "jsonSchema": {
+          "type": "string"
+        },
+        "schema": {
+          "type": "object"
```
royhadad, line 68: also here, we need to check if it's a valid json schema, in some way

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
...
+    "customRules": {
+      "type": "array",
+      "items": {
+        "identifier": "string",
+        "pattern": "^([A-Z0-9])+(_?[A-Z0-9])*$",
+        "defaultMessageOnFailure": {
+          "type":"string"
+        },
+        "jsonSchema": {
+          "type": "string"
```
royhadad, line 65: We need to check if it's a valid json schema, in some way

pkg/policy/policiesSchema.json
royhadad: what about the optional "name" key?

pkg/policy/policiesSchema.json
```diff
@@ -0,0 +1,78 @@
...
+    "apiVersion": {
+      "type": "string",
+      "enum": [
+        "v1"
+      ]
+    },
+    "message": {
+      "type": "string",
+      "enum": [
+        "Should have exact one policy set as default",
```
royhadad: where are the checks for these 3?
"Should have exact one policy set as default"
"All policy names should be unique"
"All rule identifiers should be unique"

pkg/policy/policiesSchema.json
royhadad, line 32: where is the check that validates rule identifier exists in either custom rules or default rules?

royhadad (changes requested):

pkg/policy/policiesSchema.json
Ben-Zaad: I think this is one of the abilities that we are losing due to JSON Schema limitations.

https://github.com/json-schema-org/json-schema-spec/wiki/Scope-of-JSON-Schema-Validation

pkg/policy/policiesSchema.json
Ben-Zaad: I think this is one of the abilities that we are losing due to JSON Schema limitations.

https://github.com/json-schema-org/json-schema-spec/wiki/Scope-of-JSON-Schema-Validation

dimabru (approved):

royhadad (approved):

PR #505, opened 2022-03-16
feat: add nsa rules

hadar-co:


pkg/policy/nsaHardeningRules.yaml
```diff
@@ -0,0 +1,359 @@
...
+                enum:
+                  - Pod
+          then:
+            properties:
+              spec:
+                properties:
+                  securityContext:
+                    properties:
+                      readOnlyRootFilesystem:
+                        enum:
```
eyarz, line 41: is this a string or a boolean? if it's a boolean - use const:
https://stackoverflow.com/a/57146508/9595106

pkg/policy/nsaHardeningRules.yaml
```diff
@@ -0,0 +1,359 @@
...
+        $ref: "#"
+- id: 36
+    name: Prevent containers from accessing underlying host
+    uniqueName: NSA_INCORRECT_KEY_HOSTPATH
+    enabledByDefault: true
+    documentationUrl: 'https://hub.datree.io/'
+    messageOnFailure: Invalid key `hostPath` - refrain from using this mount to prevent an attack on the underlying host  
+    category: NSA-Hardening
+    schema:
+        definitions:
```
eyarz, line 63: fix indentation

pkg/policy/nsaHardeningRules.yaml
```diff
@@ -0,0 +1,359 @@
...
+            properties:
+              containers:
+                type: array
+                items:
+                  properties:
+                    ports:
+                      type: array
+                      items:
+                        not:
+                          required:
```
eyarz, line 214: why not `cosnt` instead of `required`?

eyarz: ignore my comment, I didn't notice you are verifying the key and not the value

PR #509, opened 2022-03-20
fix: add kustomize cmd description

myishay:


cmd/kustomize/main.go
```diff
@@ -94,7 +94,8 @@ func New(testCtx *test.TestCommandContext, kustomizeCtx *KustomizeContext) *cobr
 	testCommandFlags.AddFlags(kustomizeTestCommand)
 
 	kustomizeCommand := &cobra.Command{
-		Use: "kustomize",
+		Use:   "kustomize",
+		Short: "Render kustomize configs and run a policy check against them",
```
royhadad: no long description?

hadar-co: @adifay 
I recommend using similar/identical wording in all places for the same feature.
In the docs, we wrote it like this:
Run "datree kustomize test [path]" to render files of a kustomization.yaml file located in [path] and run a policy check against them.

Here I would suggest:
Render files defined in a kustomization.yaml file and run a policy check against them

myishay: @royhadad  No need. it is only a reference to the test command

adifayer (dismissed):

hadar-co (dismissed):

cmd/kustomize/main.go
```diff
@@ -94,7 +94,8 @@ func New(testCtx *test.TestCommandContext, kustomizeCtx *KustomizeContext) *cobr
 	testCommandFlags.AddFlags(kustomizeTestCommand)
 
 	kustomizeCommand := &cobra.Command{
-		Use: "kustomize",
+		Use:   "kustomize",
+		Short: "Render files defined in a kustomization.yaml file and run a policy check against them",
```
adifayer: Render configs defined in a kustomization.yaml file and run a policy check against them

adifayer (changes requested):

hadar-co (approved):

adifayer (approved):

PR #510, opened 2022-03-20
feat: policy-config flag

dimabru:


bl/evaluation/evaluator_test.go
```diff
@@ -276,7 +276,7 @@ func newFilesConfigurations(path string) []*extractor.FileConfigurations {
 }
 
 func mockGetPreRunData() *cliClient.EvaluationPrerunDataResponse {
-	const policiesJsonPath = "../../internal/fixtures/policyAsCode/policies.json"
+	const policiesJsonPath = "../../internal/fixtures/policyAsCode/prerun.json"
```
TzlilSwimmer123, line 279: 💯

bl/policy/policy_factory.go
```diff
@@ -60,6 +68,40 @@ func CreatePolicy(policies *cliClient.EvaluationPrerunPolicies, policyName strin
...
+	fileReader := fileReader.CreateFileReader(nil)
+	policiesStr, err := fileReader.ReadFileContent(path)
+	if err != nil {
+		return nil, err
+	}
+	validationResults, err := validatePoliciesYaml(policiesStr)
+	if err != nil {
+		return nil, err
+	}
+	if !validationResults.Valid() {
```
alexfedin: This should be part of the validatepoliciesYaml function because then you have 2 places that are responsible for validation, the `validatePoliciesYaml` function and the if scope in this function

bl/policy/policy_factory.go
```diff
@@ -60,6 +68,40 @@ func CreatePolicy(policies *cliClient.EvaluationPrerunPolicies, policyName strin
 	return Policy{policyName, rules}, nil
 }
 
+func GetPoliciesFileFromPath(path string) (*cliClient.EvaluationPrerunPolicies, error) {
+	fileReader := fileReader.CreateFileReader(nil)
+	policiesStr, err := fileReader.ReadFileContent(path)
```
alexfedin: To make this function cleaner, so it will be responsible only for one thing, you should extract reading the file outside and then validate the contents of the file outside. After that you can pass the contents to `GetPoliciesFileFromPath` function so it will only be responsible for extracting the policies, as the name of the functions suggests

pkg/jsonSchemaValidator/validator.go
```diff
@@ -0,0 +1,37 @@
...
+
+	json, _ := yaml.YAMLToJSON([]byte(yamlContent))
+
+	schemaLoader := gojsonschema.NewStringLoader(string(jsonSchema))
+	documentLoader := gojsonschema.NewStringLoader(string(json))
+
+	return gojsonschema.Validate(schemaLoader, documentLoader)
+}
+
+// func (jsv *JSONSchemaValidator) Validate(jsonSchema map[string]interface{}, yamlContent string) (*Result, error) {
```
alexfedin: remove redundant comment

pkg/jsonSchemaValidator/validator.go
```diff
@@ -0,0 +1,37 @@
...
+	return &JSONSchemaValidator{}
+}
+
+type Result = gojsonschema.Result
+
+func (jsv *JSONSchemaValidator) ValidateYamlSchema(schemaContent string, yamlContent string) (*Result, error) {
+
+	jsonSchema, _ := yaml.YAMLToJSON([]byte(schemaContent))
+
+	json, _ := yaml.YAMLToJSON([]byte(yamlContent))
```
alexfedin: Use Validate function here to avoid code duplication

alexfedin (changes requested):

pkg/cliClient/cliClient_test.go
```diff
@@ -342,7 +342,7 @@ func test_getVersionMessage_success() *GetVersionMessageTestCase {
 }
```
royhadad, line 342: why did you rename policies.json to prerun.json?

dimabru: because the file is actually a mock of the response of the `prerun` call to the backend

bl/policy/policy_factory.go
dimabru: done

bl/policy/policy_factory.go
dimabru: @MentalBrake 
what if I change the name of the function to `populatePolicies(path string)`?

alexfedin: Then you still get a function that is responsible for reading a file, validating a yaml against json schema and parsing it into policies, this creates a coupling between multiple functionalities

dimabru: I prefer to keep the logic of getting the policy post-validation inside the policy_factory package. I don't mind splitting the `GetPoliciesFileFromPath` into several functions with one 'above' that will call the file creation, conversion and validation.
What should I name the function inside `policy_factory`?
I want to solve this without having to add the validation and file retrieving logic to the caller function (which is the `test` command)

alexfedin: we don't have to make these changes, it can be refactored later, anyway this function does not belong to a policy factory, the factory should only be responsible for creating objects, move this to policy package

alexfedin (dismissed):

alexfedin (approved):

PR #511, opened 2022-03-20
refactor: move argument logic to `Args` block

### LINKED ISSUE #433: Validate test command args before cobra.Run

Validate test commands arguments before Run. This pattern will allow us to governance  behaviors such as flags/arguments validation in all commands. Additionally, this should make the code easier to understand and maintain. 

**Requirements**
Golang basic level.

- `cmd/test/main.go` 
You can see that in `test()` we check if reading files from the `stdn` is necessary. Before, we check the commands args. If they’re invalid we throw an error. This issue is to move this logic and any other arguments validation(such as `Validate()` function)  to `cobra.Command` Args field.
- The final result should be something like this:
    
    ```go
    Args: func(cmd *cobra.Command, args []string) error {
    			if len(args) < 1 {
    				return fmt.Errorf("requires at least 1 arg")
    			} else if args[0] == "-" {
    				return fmt.Errorf(fmt.Sprintf("Unexpected args: [%s]", strings.Join(args[1:], ",")))
    			}
    
    			return testCommandFlags.Validate()
    }
    ```

amustaque97:
Fixes: #433 

**Changes proposed in the PR**
In this PR we need to move arguments validation logic in the `Args` block of cobra command. 

**Testing Instruction**
Since this is a refactor work make sure all test cases are passing, build is successful and functionality is working as expected.

### DIFF
--- a/cmd/kustomize/main.go
+++ b/cmd/kustomize/main.go
@@ -1,6 +1,7 @@
 package kustomize
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
@@ -43,6 +44,14 @@ func New(testCtx *test.TestCommandContext, kustomizeCtx *KustomizeContext) *cobr
 		# Test the kustomize build from github
 		datree kustomize test https://github.com/kubernetes-sigs/kustomize.git/examples/helloWorld?ref=v1.0.6
 		`),
+		Args: func(cmd *cobra.Command, args []string) error {
+			if args[0] == "-" {
+				if len(args) > 1 {
+					return fmt.Errorf(fmt.Sprintf("Unexpected args: [%s]", strings.Join(args[1:], ",")))
+				}
+			}
+			return nil
+		},
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return test.LoadVersionMessages(testCtx, args, cmd)
 		},
--- a/cmd/kustomize/main.go
+++ b/cmd/kustomize/main.go
@@ -1,7 +1,6 @@
 package kustomize
 
 import (
-	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


cmd/test/main.go
```diff
@@ -188,11 +188,6 @@ func New(ctx *TestCommandContext) *cobra.Command {
 				}
 			}()
 
-			err = testCommandFlags.Validate()
```
royhadad, line 192: Good catch! seems like we had a duplicated check

cmd/test/main.go
```diff
@@ -287,9 +282,6 @@ func validateK8sVersionFormatIfProvided(k8sVersion string) error {
 func Test(ctx *TestCommandContext, paths []string, prerunData *TestCommandData) error {
 
 	if paths[0] == "-" {
-		if len(paths) > 1 {
```
royhadad, line 304: If we move this check to `Args` then it is not tested in the kustomize call anymore https://github.com/datreeio/datree/blob/a403a0cbe12883efcd855709123799be442200f2/cmd/kustomize/main.go#L87

I think we want to keep this check here

royhadad (changes requested):

cmd/kustomize/main.go
```diff
@@ -43,6 +44,14 @@ func New(testCtx *test.TestCommandContext, kustomizeCtx *KustomizeContext) *cobr
 		# Test the kustomize build from github
 		datree kustomize test https://github.com/kubernetes-sigs/kustomize.git/examples/helloWorld?ref=v1.0.6
 		`),
+		Args: func(cmd *cobra.Command, args []string) error {
```
royhadad, line 46: this solves the problem! can we just extract this check to avoid code duplication?

royhadad (changes requested):

cmd/test/main.go
amustaque97: IMO, it would be better to introduce `Args` block in `Kustomize`. I'm expecting going forward we will gonna introduce it. Also, it will give more control individually over each command `test` and `kustomize`.

WDYT?

cmd/kustomize/main.go
amustaque97: I made the required changes and hope that changes are based on golang code standards.

amustaque97: fixing the fail stages of pipelines

royhadad (approved):

royhadad:
@amustaque97 Thank you for the contribution! Great job! 👏

PR #514, opened 2022-03-21
feat: new argo rules

adifayer:
10 new rules <3


pkg/policy/defaultRules.yaml
```diff
@@ -881,3 +881,314 @@ rules:
                       anyOf:
                         - pattern: ^(169\.254\.)
                         - pattern: ^(127\.)
+  - id: 34
```
dimabru: this is a copy of the exising rule 34 - should be deleted I think

adifayer: its an old rule, mine start from 35

adifayer: see line 910 pls

pkg/policy/defaultRules.yaml
```diff
@@ -881,3 +881,314 @@ rules:
...
+                  type: array
+                  items:
+                    not:
+                      anyOf:
+                        - pattern: ^(169\.254\.)
+                        - pattern: ^(127\.)
+  - id: 35
+    name: 'Ensure DAG fail-fast on node failure'
+    uniqueName: 'ARGO_WORKFLOW_INCORRECT_FAILFAST_VALUE_FALSE'
+    enabledByDefault: true
```
dimabru: new rules in default policies should be disabled by default
otherwise users that do not have the latest cli with local evaluation will fail because they are missing these rules' implementations

pkg/policy/defaultRules.yaml
```diff
@@ -881,3 +881,314 @@ rules:
...
+    documentationUrl: # TODO
```
dimabru: add something here to avoid potential issues when testing
you can simply remove the comment `#`

pkg/policy/defaultRules.yaml
dimabru: yeah I know. you accidently added the same rule twice

adifayer: when you're right, you're not wrong

pkg/policy/defaultRules.yaml
adifay: We should have a development task meant to support this behavior- for the new argo policy, these rules are turned on by default. For default policy- should be off. @royhadad correct?

royhadad (changes requested):
I did not review the entire pr, just requesting changes to block it from being merged.
We need to handle the error where a user has an old CLI with new rules enabled.
basically - detect the error and print a user friendly error message

dimabru (dismissed):

royhadad (dismissed):

royhadad (dismissed):

royhadad (approved):

PR #517, opened 2022-03-21
feat: add offline flag in config yaml

TzlilSwimmer123:


TzlilSwimmer123:
Need to add more tests and open for design refactor

cmd/kustomize/main.go
```diff
@@ -61,18 +62,28 @@ func New(testCtx *test.TestCommandContext, kustomizeCtx *KustomizeContext) *cobr
...
+
+			if err != nil && isBackendAvailable {
 				return err
 			}
 
-			evaluationPrerunData, err := testCtx.CliClient.RequestEvaluationPrerunData(localConfigContent.Token)
+			evaluationPrerunData := &cliClient.EvaluationPrerunDataResponse{}
 
+			if isBackendAvailable {
+				evaluationPrerunData, err = testCtx.CliClient.RequestEvaluationPrerunData(localConfigContent.Token)
```
alexfedin: it looks like the way isBackendAvailable is being used is always the same

Move the `isBackednAvailable` logic inside the clients function so that it will be exposed only to cliClient, that way when you request something and backend is down the cliClient function will return the empty value of what you request because that's what you're doing here anyway.

This will also require the offline parameter to be passed to the cliClient so it will know if it should return an error or an empty value

alexfedin (changes requested):
@TzlilSwimmer123 missing handling the reportErrors requests in main.go so they will be ignored too, in general if you will put the logic to ignore network errors inside the cliClient you won't have to worry about where it's being used and it will hold for future requests as well

dimabru:
@TzlilSwimmer123 
bump this to version 1.1.0 :)

dimabru (changes requested):
@TzlilSwimmer123
bump this to version 1.1.0 :)

alexfedin (dismissed):

dimabru (dismissed):

alexfedin (approved):

PR #522, opened 2022-03-22
test: added for extractor

### LINKED ISSUE #443: Add test coverage for extractor package

Add test coverage for package extractor. See `pkg/httpClient` as a reference.

For `ToAbsolutePath`, the test cases are already exists but for  `ExtractConfigurationsFromYamlFile` the test cases should include:

- Case: empty(””) path argument
    
    Expected: should return invalid file with returned `ToAbsolutePath` error.
    
- Case: reading file failed
    
    Expected: should return invalid file with returned error from `ReadFileContent`
    
- Case: parsing YAML failed
    
    Expected: should return invalid file with returned error from `ParseYaml`
    
- Case: success
    
    Expected: should return expected configurations, absolute path and nil
    

**“How to” implementation suggestions**

For `ToAbsolutePath`, the test cases are already exists. Use them by moving `TestToAbsolutePath` to the new `extractor_test` file (find, `TestToAbsolutePath` in `bl/files`).

For `ExtractConfigurationsFromYamlFile` tests

**Requirements**
Golang basic level.

sourikghosh:
closes #443

changes:
- relocated test-cases for ToAbsolutePath 
- added test-cases for ExtractConfigurationsFromYamlFile


sourikghosh:
review @noaabarki @MentalBrake

TzlilSwimmer123:
Thanks for your effort @sourikghosh 😃  
We will look into it in the next couple of days 🙏

pkg/extractor/extractor_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			absolutePath, _ := ToAbsolutePath(tt.args.path)
+			assert.Equal(t, tt.expected.path, absolutePath)
+		})
+	}
+}
+
+func TestExtractConfigurationsFromYamlFile(t *testing.T) {
+	t.Run("invalid yaml path, should return invalid file with returned ToAbsolutePath error", func(t *testing.T) {
+		actualVal, actualStr, actualErr := ExtractConfigurationsFromYamlFile("")
```
TzlilSwimmer123: I think that `actualVal, actualStr, actualErr` are not easy to understand what we expect the function to return, if I want to know it I need to go the function declaration. What do you think about renaming in to something more meaningful like `configurations, absolutePath, ...` ?

pkg/extractor/extractor_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+		assert.Empty(t, actualVal)
+		assert.Equal(t, "", actualStr)
```
TzlilSwimmer123: I think that it's a good practice to create new variable for empty string and use it here and when you call the function. Prevent using `""` twice

pkg/extractor/extractor_test.go
```diff
@@ -0,0 +1,102 @@
...
+		assert.NotNil(t, actualErr)
+	})
+
+	t.Run("invalid yaml file, should return an error", func(t *testing.T) {
+		path := "../../internal/fixtures/policyAsCode/invalid-yaml.yaml"
+		actualVal, actualResult, actualErr := ExtractConfigurationsFromYamlFile(path)
```
TzlilSwimmer123: Same comment about renaming

pkg/extractor/extractor_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+		assert.Empty(t, actualVal)
+		assert.Equal(t, "", actualResult)
+		assert.EqualValues(t, "yaml validation error: yaml: line 2: did not find expected key\n", actualErr.ValidationErrors[0].Error())
```
TzlilSwimmer123: I think you used `assert.EqualValues` to check if 2 objects are equal but now you check for a string value so it's redundant. You can use `Equal`

pkg/extractor/extractor_test.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+
+	t.Run("valid yaml file, should return a Configuration and no error", func(t *testing.T) {
+		path := "../../internal/fixtures/jsonSchema/yamlSchema.yaml"
+		actualValue, actualResult, actualErr := ExtractConfigurationsFromYamlFile(path)
```
TzlilSwimmer123: Same comment about renaming

pkg/extractor/extractor_test.go
```diff
@@ -0,0 +1,102 @@
...
+	t.Run("invalid yaml file, should return an error", func(t *testing.T) {
+		path := "../../internal/fixtures/policyAsCode/invalid-yaml.yaml"
+		actualVal, actualResult, actualErr := ExtractConfigurationsFromYamlFile(path)
+
+		assert.Empty(t, actualVal)
+		assert.Equal(t, "", actualResult)
+		assert.EqualValues(t, "yaml validation error: yaml: line 2: did not find expected key\n", actualErr.ValidationErrors[0].Error())
+	})
+
+	t.Run("valid yaml file, should return a Configuration and no error", func(t *testing.T) {
```
TzlilSwimmer123: I think it's better to align the capitalization and put `configuration`

TzlilSwimmer123 (changes requested):
Hi @sourikghosh,
Really appreciate your time and work.

Left some comments :)

TzlilSwimmer123 (approved):
Looks good!!!
Thanks

PR #525, opened 2022-03-23
fix: publishing custom rules using jsonSchema fails in datree test

### LINKED ISSUE #523: publishing custom rules using jsonSchema (instead of schema) fails in `datree test`
**Describe the bug**
publishing custom rules using the jsonSchema key (instead of schema) fails later in `datree test`

**To Reproduce**
1. Run command `datree publish ./internal/fixtures/policyAsCode/valid-schema.yaml`
2. Go to the dashboard => policies => default policy - turn on the rule with the name "rule unique 2"
3. Run command `datree test ~/.datree/k8s-demo.yaml`
4. See error "schema is invalid"

**Expected behavior**
The custom rule "rule unique 2" should run normally

**Screenshots**
Screen Shot 2022-03-22 at 18 48 20

 - OS: Mac

 - Version: 1.0.6

**Additional context**  
This happens because the code expects EvaluationPrerunPolicies.CustomRules.schema to exist.
In reality, either the "schema" key exists OR the "jsonSchema" key exists - we need to parse both and choose the existing option.
Screen Shot 2022-03-22 at 18 50 23

Ben-Zaad:


alexfedin (approved):

bl/policy/policy_factory.go
```diff
@@ -72,7 +73,16 @@ func populateRules(policyRules []cliClient.Rule, customRules []*cliClient.Custom
 		customRule := getCustomRuleByIdentifier(customRules, rule.Identifier)
 
 		if customRule != nil {
-			rules = append(rules, RuleWithSchema{rule.Identifier, customRule.Name, customRule.Schema, rule.MessageOnFailure})
+			if customRule.Schema == nil {
+				schema := make(map[string]interface{})
+				err := json.Unmarshal([]byte(customRule.JsonSchema), &schema)
```
royhadad, line 78: please consider extracting this into a function, I think we use this several times.
`func jsonSchemaToStruct()`

we already have `func yamlToStruct()`

royhadad (approved):

bl/policy/policy_factory.go
alexfedin: 1. `yamlToStruct` is not used in a general way, it's a function that is used with a specific struct so I don't even know why it's called `yamlToStruct`. 
2. Even though there is a yamlToStruct function we still use `yaml.Unmarshall` outside of this function, so that indicates on what happens with utils function in general so wrapping it in `jsonSchemaToStruct` function will not guarantee usage of this function in the future

alexfedin (approved):

PR #526, opened 2022-03-23
fix: handle config creation error if insufficient permissions

myishay:


dimabru (changes requested):
I don't think we should ignore all errors. can we be more specific and ignore only the one(s) that is causing the bug?

myishay:
> I don't think we should ignore all errors. can we be more specific and ignore only the one(s) that is causing the bug?

I am only ignoring the ones that are causing the bug

dimabru (approved):

PR #530, opened 2022-03-28
fix: use correct user homedir

myishay:
* Uses correct user homedir following this comment https://github.com/Homebrew/homebrew-core/pull/97061#issuecomment-1078690339
* Reverts changes made in https://github.com/datreeio/datree/pull/526
* Added a helper function for duplicate code


pkg/localConfig/localConfig.go
```diff
@@ -58,12 +57,12 @@ func (lc *LocalConfigClient) GetLocalConfiguration() (*LocalConfig, error) {
...
 	if offline == "" {
-		viper.SetDefault(offlineKey, "fail")
-		_ = viper.WriteConfig()
-		_ = viper.ReadInConfig()
-		offline = viper.GetString(offlineKey)
+		offline = "fail"
+		err := setViperVariable(offlineKey, offline)
+		if err != nil {
+			return nil, err
+		}
```
TzlilSwimmer123, line 64: Let's make sure we don't need to `return &LocalConfig{}, err` instead of `return nil, err`

pkg/localConfig/localConfig.go
```diff
@@ -72,17 +71,18 @@ func (lc *LocalConfigClient) GetLocalConfiguration() (*LocalConfig, error) {
...
 		}
 		token = createTokenResponse.Token
-		viper.SetDefault(tokenKey, token)
-		_ = viper.WriteConfig()
-		_ = viper.ReadInConfig()
-		token = viper.GetString(tokenKey)
+		err = setViperVariable(tokenKey, token)
+		if err != nil {
+			return nil, err
+		}
```
TzlilSwimmer123, line 80: same as above

pkg/localConfig/localConfig.go
```diff
@@ -72,17 +71,18 @@ func (lc *LocalConfigClient) GetLocalConfiguration() (*LocalConfig, error) {
...
 	if clientId == "" {
-		viper.SetDefault(clientIdKey, shortuuid.New())
-		_ = viper.WriteConfig()
-		_ = viper.ReadInConfig()
-		clientId = viper.GetString(clientIdKey)
+		clientId = shortuuid.New()
+		err := setViperVariable(clientIdKey, clientId)
+		if err != nil {
+			return nil, err
+		}
```
TzlilSwimmer123, line 88: same as above

pkg/localConfig/localConfig.go
```diff
@@ -58,31 +57,35 @@ func (lc *LocalConfigClient) GetLocalConfiguration() (*LocalConfig, error) {
 	offline := viper.GetString(offlineKey)
 
 	if offline == "" {
-		viper.SetDefault(offlineKey, "fail")
-		_ = viper.WriteConfig()
-		_ = viper.ReadInConfig()
-		offline = viper.GetString(offlineKey)
+		offline = "fail"
```
shmuelsa, line 60: `fail` can set on they global const 
`failKey          = "fail"`

your call

pkg/localConfig/localConfig.go
```diff
@@ -180,6 +200,25 @@ func setViperConfig() (string, string, string, error) {
 	return configHome, configName, configType, nil
 }
 
+func setViperVariable(key string, value string) error {
```
shmuelsa, line 203: NICE!

shmuelsa (approved):

PR #531, opened 2022-03-28
feat: skip k8s validation when no connection

dimabru:


bl/evaluation/printer.go
```diff
@@ -21,7 +22,7 @@ type Printer interface {
 	PrintEvaluationSummary(summary printer.EvaluationSummary, k8sVersion string)
 }
 
-func PrintResults(results FormattedResults, invalidYamlFiles []*extractor.InvalidFile, invalidK8sFiles []*extractor.InvalidFile, evaluationSummary printer.EvaluationSummary, loginURL string, outputFormat string, printer Printer, k8sVersion string, policyName string) error {
+func PrintResults(results FormattedResults, invalidYamlFiles []*extractor.InvalidFile, invalidK8sFiles []*extractor.InvalidFile, evaluationSummary printer.EvaluationSummary, loginURL string, outputFormat string, printer Printer, k8sVersion string, policyName string, k8sValidationWarnings validation.K8sValidationWarningPerValidFile) error {
```
alexfedin: I think that 10 parameters is enough to finally aggregate the function input into it's own variable

bl/evaluation/printer.go
```diff
@@ -49,7 +50,7 @@ func PrintResults(results FormattedResults, invalidYamlFiles []*extractor.Invali
 			return xmlOutput(&formattedOutput)
 		}
 	} else {
-		return textOutput(results.EvaluationResults, invalidYamlFiles, invalidK8sFiles, evaluationSummary, loginURL, printer, k8sVersion, policyName)
+		return textOutput(results.EvaluationResults, invalidYamlFiles, invalidK8sFiles, evaluationSummary, loginURL, printer, k8sVersion, policyName, k8sValidationWarnings)
```
alexfedin: same goes for this function

bl/validation/k8sValidator_test.go
```diff
@@ -46,7 +46,7 @@ func test_valid_multiple_configurations(t *testing.T) {
 		Configurations: []extractor.Configuration{},
```
alexfedin, line 48: missing test for warning when there is no network

cmd/test/validationManager.go
```diff
@@ -63,6 +85,14 @@ func (v *ValidationManager) ValidK8sConfigurationsCount() int {
 	return totalConfigs
 }
 
+func (v *ValidationManager) SaveK8sValidationWarningPerValidFile(k8sValidationWarningPerValidFile *validation.K8sValidationWarningPerValidFile) {
```
alexfedin: Save has an implicit meaning of persistence, it is more appropriate to call this `SetK8sValidationWarningPerValidFile`

pkg/printer/printer.go
```diff
@@ -95,6 +97,14 @@ func (p *Printer) printK8sValidationWarning(warning Warning) {
 	fmt.Fprintln(out)
 }
 
+func (p *Printer) printK8sValidationWarning(warning Warning) {
+	p.printInColor("[?] Kubernetes schema validation\n", p.Theme.Colors.White)
```
alexfedin: There is a problem with printing white on white background using this function, talk to @TzlilSwimmer123 she has already found a solution for it.

alexfedin (changes requested):

bl/validation/k8sValidator.go
```diff
@@ -56,7 +60,7 @@ func (val *K8sValidator) ValidateResources(filesConfigurationsChan chan *extract
 			}
 		}
 	}()
-	return validK8sFilesConfigurationsChan, invalidK8sFilesChan
+	return validK8sFilesConfigurationsChan, invalidK8sFilesChan, k8sValidationWarningPerValidFile
```
alexfedin: You return this value before it's being written to, so potentially you can read it outside the function before it's being populated, either return this value in a channel or initialize it in the same goroutine that returns the value

bl/validation/k8sValidator.go
```diff
@@ -38,16 +41,17 @@ func (val *K8sValidator) ValidateResources(filesConfigurationsChan chan *extract
 
 		for fileConfigurations := range filesConfigurationsChan {
 
-			isValid, validationErrors, err := val.validateResource(fileConfigurations.FileName)
+			isValid, validationErrors, err, validationWarning := val.validateResource(fileConfigurations.FileName)
```
alexfedin: I think you should return the error object first or last, though I can't find any reference for that practice I think it makes the usage of errors more organized

bl/validation/k8sValidator.go
```diff
@@ -94,29 +98,37 @@ func (val *K8sValidator) isK8sFile(fileConfigurations []extractor.Configuration)
...
-func (val *K8sValidator) validateResource(filepath string) (bool, []error, error) {
+func (val *K8sValidator) validateResource(filepath string) (bool, []error, error, *ValidationWarning) {
 	f, err := os.Open(filepath)
 	if err != nil {
-		return false, []error{}, fmt.Errorf("failed opening %s: %s", filepath, &InvalidK8sSchemaError{ErrorMessage: err.Error()})
+		return false, []error{}, fmt.Errorf("failed opening %s: %s", filepath, &InvalidK8sSchemaError{ErrorMessage: err.Error()}), nil
 	}
-	defer f.Close()
+
+	defer func() {
```
alexfedin: redundant function, remove it and just use defer.Close()

bl/validation/main.go
```diff
@@ -5,6 +5,10 @@ import (
 	"strings"
 )
 
+type ValidationWarning struct {
```
alexfedin: why use a struct with only 1 field, just do var ValidationWarning string

alexfedin (changes requested):

bl/validation/main.go
dimabru: easier and more intuitive to check for `nil` value if no warning message (which is optional)

alexfedin: if `validationWarning.errorMesage != nil` is the same as checking if errorMessage != ""

but the addition of the struct adds more information to the code, no if I see it I may assume that there is something more to validationWarning other than the message and I have to go to the struct implementation to check it, whereas with a simple string I know immediately what it is

bl/validation/k8sValidator.go
alexfedin: @dimabru btw, found it, https://gobyexample.com/errors by convention errors are the last return value

bl/validation/k8sValidator.go
```diff
@@ -50,7 +57,12 @@ func (val *K8sValidator) ValidateResources(filesConfigurationsChan chan *extract
...
 			if isValid {
 				validK8sFilesConfigurationsChan <- fileConfigurations
-				k8sValidationWarningPerValidFile[fileConfigurations.FileName] = validationWarning
+				if validationWarning != nil {
+					k8sValidationWarningPerValidFileChan <- &FileWithWarning{
+						Filename: fileConfigurations.FileName,
+						Warning:  validationWarning,
+					}
+				}
 				_ = validationWarning
```
alexfedin: remove development leftovers

alexfedin (changes requested):

bl/validation/main.go
dimabru: `string` does not have a nil value. checking for `errorMessage != ""` is not the same as `!= nil`
I use a check for existing warning message in other places in the code and I think the `validationWarning` object should be treated same as `error` as far as checking for existance goes.

however, I think this discussion is taking too long so I will make the change to advance the pr

alexfedin (approved):

PR #532, opened 2022-03-28
feat: add pass/fail yamls for nsa rules

hadar-co:


royhadad (changes requested):
As indicated in the screenshot, 5 of the 18 test files don't pass k8s validation
you may reproduce this by running:
`./datree test ./pkg/policy/nsaHardeningRules/tests/*.yaml`
It's important that the test files are valid for testing purposes 😊 

Screen Shot 2022-03-30 at 18 31 40

royhadad (dismissed):

royhadad (approved):

PR #547, opened 2022-04-02
fix: adjust error message when no args given

### LINKED ISSUE #543: Running `datree kustomize test` without a path throws an unexpected error
**Describe the bug**
Running `datree kustomize test` without a path throws an unexpected error

**To Reproduce**
1. Run command `datree kustomize test`
2. see the unexpected error thrown "Unexpected error: runtime error: index out of range [0] with length 0"

**Expected behavior**
Should get the correct error message: "Requires at least 1 arg"

**Screenshots**
Screen Shot 2022-03-30 at 18 59 55

 - OS: MacOS

 - Version: 1.1.6 (this bug exists since 1.0.12-rc)
 
 **Steps to solve this bug:**
 inside pkg/utils/utils.go ValidateStdinPathArgument add the following check:
`if len(args) < 1 {
	return fmt.Errorf("Requires at least 1 arg\n")
}`

and remove the now redundant check in cmd/test/main.go line 169

naorpeled:
Closes issue #543

### DIFF
--- a/cmd/test/main.go
+++ b/cmd/test/main.go
@@ -169,10 +169,6 @@ func New(ctx *TestCommandContext) *cobra.Command {
 		cat kube-prod/deployment.yaml | datree test -
 		`),
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) < 1 {
-				errMessage := "Requires at least 1 arg\n"
-				return fmt.Errorf(errMessage)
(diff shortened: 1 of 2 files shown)


royhadad (approved):

royhadad:
Thank you for the contribution! 😊

PR #561, opened 2022-04-07
feat: add metadata.name rule

hadar-co:
@royhadad Please review but do not merge
@adifay Let's sync when this can be merged


pkg/policy/defaultRules.yaml
```diff
@@ -1170,3 +1170,29 @@ rules:
                       required:
                         - retryPolicy
                         - expression
+  - id: 45
+    name: Ensure resource has a configured name 
+    uniqueName: RESOURCE_MISSING_NAME
+    enabledByDefault: true
+    documentationUrl: 'https://hub.datree.io/'
```
royhadad, line 1177: missing documentation url (you're probably aware)

pkg/policy/nsaHardeningRules/nsaHardeningRules.yaml
```diff
@@ -1,4 +1,4 @@
-- id: 45
+- id: 46
```
royhadad, line 1: I think you'll need to also change the ids of the test files

pkg/policy/defaultRules.yaml
```diff
@@ -1170,3 +1170,29 @@ rules:
...
+    messageOnFailure: 'Missing key `name` - must be set to apply resource to cluster'
+    category: Other
+    schema:
```
royhadad, line 1180: this schema gives a false-positive when supplied with the metadata key, without any properties inside, added a screenshot for reference:

Screen Shot 2022-04-09 at 20 52 56

I think we have this problem in several other rules, so it might be a good idea to look into that.

For this rule, the simple fix I found is simply adding ```type: object``` for metadata (in both schemas) - seems to fix the problem :)

royhadad (dismissed):

Mitumaru (dismissed):

PR #567, opened 2022-04-11
fix: wrong failed rule counter

shalev007:


pkg/printer/printer.go
```diff
@@ -176,12 +176,14 @@ func (p *Printer) PrintWarnings(warnings []Warning) {
 
 			for _, failedRule := range warning.FailedRules {
 				var occurrencesPostfix string
-				if failedRule.Occurrences > 1 {
+				occurrencesCount := failedRule.Occurrences
+
+				if occurrencesCount > 1 {
```
myishay: no need to change the code here

bl/evaluation/printer.go
```diff
@@ -236,6 +235,7 @@ func parseToPrinterWarnings(results *EvaluationResults, invalidYamlFiles []*extr
 						})
 					} else {
 						hasFailedOccurrences = true
+						failedRule.Occurrences += occurrenceDetails.Occurrences
```
myishay: seems very confusing to me. I think putting the occurrences count in a method of it's own will be more readable and easier to understand

myishay (approved):

myishay (approved):

PR #575, opened 2022-04-14
fix: allow offline: local mode to run without token

myishay:


TzlilSwimmer123 (approved):
👍

pkg/localConfig/localConfig.go
```diff
@@ -71,12 +71,11 @@ func (lc *LocalConfigClient) GetLocalConfiguration() (*LocalConfig, error) {
...
-			return nil, nil
-		}
-		err = setViperVariable(tokenKey, token)
-		if err != nil {
-			return nil, err
+		if token != "" {
+			err = setViperVariable(tokenKey, token)
+			if err != nil {
+				return nil, err
+			}
```
dimabru, line 78: Your mama is so stupid, only she can understand this code

PR #577, opened 2022-04-18
feat: implement resource quotas custom keys

TzlilSwimmer123:


pkg/printer/printer.go
```diff
@@ -107,17 +109,12 @@ func (p *Printer) printK8sValidationWarning(warning Warning) {
...
-	if result.Errors() != nil {
+func (p *Printer) PrintYamlSchemaResults(errorsResult []jsonschema.Detailed, error error) {
+	if errorsResult != nil {
 		p.printInColor("Input does NOT pass validation against schema\n", p.Theme.Colors.RedBold)
 		var errorsAsString = ""
-		for _, desc := range result.Errors() {
-			fmt.Printf("- %s\n", desc)
-			errorsAsString = errorsAsString + desc.String() + "\n"
+		for _, desc := range errorsResult {
+			errorsAsString = errorsAsString + desc.InstanceLocation + " - " + desc.Error + "\n"
```
royhadad, line 117: maybe changing " - " to ": " will be less confusing in terms of UX, I don't know

pkg/jsonSchemaValidator/validator.go
```diff
@@ -12,18 +19,204 @@ func New() *JSONSchemaValidator {
...
+	m, err = toStringKeys(m)
+	if err != nil {
+		panic(err)
+	}
+
+	compiler := jsonschema.NewCompiler()
+	compiler.AssertFormat = true
+
+	if err := compiler.AddResource("schema.json", strings.NewReader(schemaContent)); err != nil {
+		panic(err)
```
royhadad: I don't think we want to panic here, this will alert pagerduty in case the user passed in an invalid yaml file!
perhaps return an error instead and handle it in the parent functions?

pkg/jsonSchemaValidator/validator.go
```diff
@@ -12,18 +19,204 @@ func New() *JSONSchemaValidator {
...
+		}
+	}
+}`)
+
+	compiler.RegisterExtension("resourceMinimum", resourceMinimum, resourceMinimumCompiler{})
+	compiler.RegisterExtension("resourceMaximum", resourceMaximum, resourceMaximumCompiler{})
+
+	schema, err := compiler.Compile("schema.json")
+	if err != nil {
+		panic(err)
```
royhadad: I don't think we want to panic here, this will alert pagerduty in case the user passed in an invalid yaml schema!
perhaps return an error instead and handle it in the parent functions?
(notice the difference between the 2 comments: yaml vs yaml-schema)

internal/fixtures/policyAsCode/custom-keys/pass-yaml-file.yaml
```diff
@@ -0,0 +1,15 @@
...
+  containers:
+    - name: cpu-demo
+      image: images.my-company.example/app:v4
+      resources:
+        requests:
+          memory: "64Mi"
+          cpu: "250m"
+        limits:
+          memory: "128Mi"
+          cpu: "350m"
```
royhadad: new line at EOF

internal/fixtures/policyAsCode/custom-keys/fail-yaml-file.yaml
```diff
@@ -0,0 +1,15 @@
...
+  containers:
+    - name: cpu-demo
+      image: images.my-company.example/app:v4
+      resources:
+        requests:
+          memory: "64Mi"
+          cpu: "250m"
+        limits:
+          memory: "128Mi"
+          cpu: "1G"
```
royhadad: new line at EOF

pkg/printer/printer.go
```diff
@@ -107,17 +109,12 @@ func (p *Printer) printK8sValidationWarning(warning Warning) {
...
 }
 
-func (p *Printer) PrintYamlSchemaResults(result *JSONSchemaValidatorResults, error error) {
-	if result == nil {
-		p.printInColor("INVALID FILE PATH\n", p.Theme.Colors.RedBold)
-		return
-	}
-	if result.Errors() != nil {
+func (p *Printer) PrintYamlSchemaResults(errorsResult []jsonschema.Detailed, error error) {
+	if errorsResult != nil {
```
royhadad, line 113: It took me a while to realize there is an early return in this if, maybe changing to `if{}...else{}` will be more readable

pkg/jsonSchemaValidator/validator.go
royhadad: perhaps return an error instead and handle it in the parent functions?

pkg/jsonSchemaValidator/validator.go
royhadad: perhaps return an error instead and handle it in the parent functions?

pkg/jsonSchemaValidator/validator_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	jsonSchemaValidator := New()
+
+	resourceYaml, _ := yaml.JSONToYAML([]byte(failResourceYamlFileContent))
+	customRuleYaml, _ := yaml.JSONToYAML([]byte(customRuleSchemaYamlFileContent))
+
+	errorsResult, _ := jsonSchemaValidator.ValidateYamlSchema(string(customRuleYaml), string(resourceYaml))
+
+	assert.GreaterOrEqual(t, len(errorsResult), 1)
```
royhadad, line 29: can we assert that errorsResult[0].Error is equal to the error string? up to you

royhadad (changes requested):
I reviewed everything accept `pkg/jsonSchemaValidator/validator.go`, let's go over it together

internal/fixtures/policyAsCode/custom-keys/custom-rule.yaml
```diff
@@ -0,0 +1,13 @@
+properties:
```
dimabru: the file name is a bit confusing. this is not a custom-rule, this is a yaml schema

royhadad (dismissed):

dimabru (dismissed):

dimabru (dismissed):

dimabru (approved):

PR #579, opened 2022-04-18
test: add a suite to validate pkg/policy/defaultRules.yaml

### LINKED ISSUE #533: Test defaultRules.yaml file structure
We have a file pkg/policy/defaultRules.yaml - which is the source of truth for the default rules in users' policies.
It's very important we don't break it, therefore - better testing is required.
A quick win would be to test and validate the structure of the yaml

Add a test file **pkg/policy/defaultRules_test.go** to test the structure of the yaml file.
**Few options here:**
1. Use a yaml-schema - pkg/jsonSchemaValidator/validator.go `ValidateYamlSchema()`
2. Convert it to json and validate using json schema
3. Parse it into a struct and validate it using code
4. A combination of both methods

**What do we want to validate?**
1. the yaml is a non-empty, non-null array
2. validate that the array contains all id's from 1 to 44 (we need this for backwards compatibility)
3. for every item in the array, it is a non-null object with the following keys:
5. id: integer, unique, bigger then 0
6. name: string, unique, not empty
7. uniqueName: string, non-empty, contains only Alphanumerical characters and underscores
8. enabledByDefault: boolean
9. documentationUrl: non empty string
10. messageOnFailure: non empty string
11. category: one of the following: 
`'Containers',
'Workload',
'CronJob',
'Networking',
'Deprecation',
'Security',
'Other',
'Custom',
'Argo'`
12. schema: non-null object. (optional - validate that it's a valid json-schema)

**Describe alternatives you've considered**
Open for suggestions for changes/additions

naorpeled:
closes #533


TzlilSwimmer123:
Thanks for your time @naorpeled  😸 
We will look into it in the next couple of days 🙏

pkg/policy/defaultRulesSchema.json
```diff
@@ -0,0 +1,65 @@
...
+        "documentationUrl",
+        "messageOnFailure",
+        "category",
+        "schema"
+      ]
+    }
+  },
+  "required": [
+    "apiVersion",
+    "rules"
```
TzlilSwimmer123, line 76: The validation uses this `required` property?

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+
+	"github.com/datreeio/datree/pkg/fileReader"
+	"github.com/datreeio/datree/pkg/jsonSchemaValidator"
+	"github.com/ghodss/yaml"
+	"github.com/stretchr/testify/assert"
+)
+
+//go:embed defaultRulesSchema.json
+var defaultRulesSchemaContent string
```
TzlilSwimmer123: The `go:embed` is used to contain not go files in go build process. There is a reason to embed it instead of reading it's content like you did in the rest of the file?

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+			validationErrors = fmt.Errorf("%s\n%s", validationErrors, validationError)
+		}
+
+		return validationErrors
+	}
+
+	return nil
+}
+
+func convertYamlFileContentToJSON(yamlFileContent string) (map[string][]interface{}, error) {
```
TzlilSwimmer123: What do you think about `convertYamlFileToUnknownStruct`? I feel like `map[string][]interface{}` doesn't 
necessarily represent a JSON

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+	yamlFileContentRawJSON, yamlParseError := yaml.YAMLToJSON([]byte(yamlFileContent))
+
+	if yamlParseError != nil {
+		return map[string][]interface{}{}, yamlParseError
```
TzlilSwimmer123: Why not returning `nil` instead of `map[string][]interface{}{}`?

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	var yamlFileContentJSON map[string][]interface{}
+	jsonMarshallingError := json.Unmarshal(yamlFileContentRawJSON, &yamlFileContentJSON)
+
+	var jsonMarshallingFailed = jsonMarshallingError != nil && reflect.TypeOf(yamlFileContentJSON) != reflect.TypeOf(map[string][]interface{}{})
+	if jsonMarshallingFailed {
+		return map[string][]interface{}{}, jsonMarshallingError
```
TzlilSwimmer123: Same question as above

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if yamlParseError != nil {
+		return map[string][]interface{}{}, yamlParseError
+	}
+
+	var yamlFileContentJSON map[string][]interface{}
+	jsonMarshallingError := json.Unmarshal(yamlFileContentRawJSON, &yamlFileContentJSON)
+
+	var jsonMarshallingFailed = jsonMarshallingError != nil && reflect.TypeOf(yamlFileContentJSON) != reflect.TypeOf(map[string][]interface{}{})
+	if jsonMarshallingFailed {
```
TzlilSwimmer123: I would love if you can explain the reason for `reflect.TypeOf(yamlFileContentJSON) != reflect.TypeOf(map[string][]interface{}{}` 🙏

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func TestDefaultRulesFileFitsJSONSchema(t *testing.T) {
+	validationError := validateYamlUsingJSONSchema(defaultRulesYamlPath, defaultRulesSchemaContent)
+
+	assert.Nil(t, validationError)
+}
+
+func TestDefaultRulesHasUniqueNamesInRules(t *testing.T) {
+	defaultRulesFileContent, _ := getFileFromPath(defaultRulesYamlPath)
```
TzlilSwimmer123, line 34: I am not sure what is the best solution but I think that we read the same file 3 / more times in this test, maybe we can read it once and use it's content in all Test functions

TzlilSwimmer123: I am not sure what is the best solution but I think that we read the same file 3 / more times in this test, maybe we can read it once and use it's content in all Test functions

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+		for _, validationError := range schemaValidationResult.Errors() {
+			validationErrors = fmt.Errorf("%s\n%s", validationErrors, validationError)
+		}
+
+		return validationErrors
+	}
+
+	return nil
+}
```
TzlilSwimmer123: 💯  
we are about to change a bit our validation mechanism so we will need to mini refactor this function. Just FYI :)

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return yamlFileContentJSON, nil
+}
+
+func validateUniqueStringPropertyValuesInArray(propertyName string, array []interface{}) error {
+	propertyValuesExistenceMap := make(map[string]bool)
+
+	for _, item := range array {
+		itemObject := item.(map[string]interface{})
+		propertyValue := itemObject[propertyName].(string)
```
TzlilSwimmer123: An interesting thing here is that `itemObject[propertyName].(string)` can return a second argument that represents the fact that if the type assertion was successful. Maybe it's good also to check if the type assertion is valid

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+		if propertyValuesExistenceMap[propertyValue] {
+			return fmt.Errorf("Property %s has duplicate value %s", propertyName, propertyValue)
```
TzlilSwimmer123: Goland is saying that you need to change `Property` to `property` 😆

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return nil
+}
+
+func validateUniqueFloat64PropertyValuesInArray(propertyName string, array []interface{}) error {
+	propertyValuesExistenceMap := make(map[float64]bool)
+
+	for _, item := range array {
+		itemObject := item.(map[string]interface{})
+		propertyValue := itemObject[propertyName].(float64)
```
TzlilSwimmer123: Same as my type assertion comment above

TzlilSwimmer123 (changes requested):
Wow, enjoyed every minute of reviewing. 
Left couple of questions and suggestions.
Please also rebase from main :)

Thanks a lot 🚀

pkg/policy/defaultRules_test.go
TzlilSwimmer123: I just tried to mess up with defaultRules.yaml and fail validation but it passes - I tried to put a problematic unique name and it passed, also minLength / maxLength doesn't fail the validation. Can you plz check it? 😺

TzlilSwimmer123 (changes requested):
For some reason the yaml validation against the json doesn't work. Can you plz check it? 🙏

pkg/policy/defaultRulesSchema.json
naorpeled: Yes,
also made sure that if those fields are missing, an error is thrown when running the test

pkg/policy/defaultRules_test.go
naorpeled: Honestly,
I'm not sure why I felt it made more sense when I wrote it.

When I'm reading it now nil makes much more sense, it allows easier assertions and a tiny bit more readable

pkg/policy/defaultRules_test.go
naorpeled: The subject of this test suite is defaultRules.go,
so I thought it made more sense to put more focus on the actual subject and embed the schema's content.

If we want to test that the schema exists and valid, we can make another test suite in my opinion, what do you think?

TzlilSwimmer123: My only concern was that the embedded file will get into our build but I checked and `*_test.go` are excluded during build time. We can keep it embedded

TzlilSwimmer123:
@naorpeled still WIP? There are some comments that I see you didn't respond on

pkg/policy/defaultRules_test.go
naorpeled: > @naorpeled still WIP? There are some comments that I see you didn't respond on

Yes,
sent you a message on Slack.

Will continue looking at the other comments when I have time (hopefully today)

pkg/policy/defaultRules_test.go
naorpeled: I’m not sure what’s the best practice regarding tests setup in Go.

In JavaScript I would use beforeAll which is a hook that is called before all the tests are run.

I saw that the testing library exposes a function called TestMain but wasn’t sure wether it was the right approach

pkg/policy/defaultRules_test.go
naorpeled: > @naorpeled still WIP? There are some comments that I see you didn't respond on

Yes.
Didn't have enough time yesterday,
also messaged you on Slack BTW

naorpeled:
> @naorpeled still WIP? There are some comments that I see you didn't respond on

Yes.
Didn't have enough time yesterday,
I replied to the rest of the comments today, will look tomorrow into why the CI is failing.

I messaged you on Slack BTW

pkg/policy/defaultRules_test.go
naorpeled: > What do you think about `convertYamlFileToUnknownStruct`? I feel like `map[string][]interface{}` doesn't necessarily represent a JSON

What about `convertYamlFileToMap`?

pkg/policy/defaultRules_test.go
naorpeled: If you don't provide the second argument Go panics, so it's kinda redundant as far as I know

pkg/policy/defaultRules_test.go
naorpeled: The JSON unmarshalling threw the following error,
```
 expected nil, but got: &json.UnmarshalTypeError{Value:"strin
g", Type:(*reflect.rtype)(0x13f3b00), Offset:139, Struct:"", Field:""}
```
I added the extra condition as a hack to prevent Golang from panicking when the unmarshalling does a good job and manages to convert the input into a map.
The tests worked with that code so I let it be when I worked on it,
definitely not proud of it in any way.

Would love to get some help with solving it in a different way 🙏

pkg/policy/defaultRules_test.go
TzlilSwimmer123: Cool

pkg/policy/defaultRules_test.go
TzlilSwimmer123: @dimabru you had a case when you used something like `beforeAll ` on our tests, where it was?

pkg/policy/defaultRules_test.go
TzlilSwimmer123: Yes, right. Well if you are ok with panic here I am also ok with it

pkg/policy/defaultRules_test.go
TzlilSwimmer123: Ok I think I was able to understand what is the best here.
First of all, the `aliases` in `defaultRules.yaml` is a bit problematic when reading the file and Unmarshaling it to json. We use an alias there and for some reason it makes problems.

I suggest the following, tell me what do you think:
- create a struct that represents the `deafultRules.yaml`:

```
type DefaultRules struct {
	ApiVersion string                  `json:"apiVersion"`
	Rules      []DefaultRuleDefinition `json:"rules"`
}
```

- When you Unmarshal you can do like this:
```
	var yamlFileContentJSON DefaultRules
	jsonUnmarshallingError := json.Unmarshal(yamlFileContentRawJSON, &yamlFileContentJSON)
```

And instead of using `defaultRulesFileContentJSON["rules"]` you can pass `defaultRulesFileContentJSON.Rules`.
In each place you can use the new type `DefaultRules`.

Example for a refactored function:
```
func validateUniqueStringPropertyValuesInArray(propertyName string, defaultRules []DefaultRuleDefinition) error {
	propertyValuesExistenceMap := make(map[string]bool)

	for _, item := range defaultRules {
		if propertyValuesExistenceMap[item.UniqueName] {
			return fmt.Errorf("property %s has duplicate value %s", propertyName, item.UniqueName)
		}

		propertyValuesExistenceMap[item.UniqueName] = true
	}

	return nil
}
```

I know you tried to make the function as generic as they can be but I think that in this case it's ok to make them especially for the validation of the file.

This way we don't need the reflect code and the code that Unmarshal to an unknown map.

Let me know if it was clear enough

TzlilSwimmer123:
Hi @naorpeled, I left a quite big comment that can help us move on quite fast with this pr. Yesterday we changed the way we validate so I thought it would be nice if I update the validation code for you. I pushed the changes and it works.

pkg/policy/defaultRules_test.go
dimabru: @TzlilSwimmer123 look at the `setup()` function in `main_test.go` file

TzlilSwimmer123: @naorpeled Basically you can test `validateUniqueStringPropertyValuesInArray` and `validateUniqueFloat64PropertyValuesInArray` in the same TestSuite, read the file in the begging and use its content in the places you need. In general I think it's good to put them in the same suite but your call

naorpeled:
Hey @TzlilSwimmer123,
updated the PR according to your comments.

TzlilSwimmer123:
Hello @naorpeled will review it :)
Thanks!

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+		propertyValuesExistenceMap[propertyValue] = true
+	}
+
+	return nil
+}
+
+func getStringValueOfRuleProperty(rule DefaultRuleDefinition, propertyName string) string {
+	itemValue := reflect.ValueOf(rule)
+	propertyValue := reflect.Indirect(itemValue).FieldByName(propertyName).String()
```
TzlilSwimmer123: I really appreciate that you tried to do this generic but I don't see the point here, we just need to make sure item.UniqueName is really unique. I think this function is redundant(+ we don't need to install a new pkg for that). See my suggestion in the father function that uses this.

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return yamlFileContentJSON, nil
+}
+
+func validateUniqueStringValuesInRulesForProperty(propertyName string, rules []DefaultRuleDefinition) error {
+	propertyValuesExistenceMap := make(map[string]bool)
+
+	for _, item := range rules {
+		propertyValue := getStringValueOfRuleProperty(item, propertyName)
```
TzlilSwimmer123: Let's change it to `propertyValue := item.UniqueName`

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+		propertyValuesExistenceMap[propertyValue] = true
+	}
+
+	return nil
+}
+
+func getIntValueOfRuleProperty(rule DefaultRuleDefinition, propertyName string) int64 {
+	itemValue := reflect.ValueOf(rule)
+	propertyValue := reflect.Indirect(itemValue).FieldByName(propertyName).Int()
```
TzlilSwimmer123: Same as the string functions - can be removed

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+func getStringValueOfRuleProperty(rule DefaultRuleDefinition, propertyName string) string {
+	itemValue := reflect.ValueOf(rule)
+	propertyValue := reflect.Indirect(itemValue).FieldByName(propertyName).String()
+
+	return propertyValue
+}
+
+func validateUniqueIntValuesInRulesForProperty(propertyName string, rules []DefaultRuleDefinition) error {
+	propertyValuesExistenceMap := make(map[int64]bool)
```
TzlilSwimmer123: Let's change it to `make(map[int]bool`

pkg/policy/defaultRules_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	for _, item := range rules {
+		propertyValue := getIntValueOfRuleProperty(item, propertyName)
```
TzlilSwimmer123: Let's change it to `item.ID`

TzlilSwimmer123 (changes requested):
Overall looks amazing!

Requested 2 small changes 🙏 

Again, thanks :)

pkg/policy/defaultRules_test.go
naorpeled: Fair points.

Thought that it would be nice to have for future scenarios where we would want to prevent more duplications, it's premature though.

naorpeled:
@TzlilSwimmer123 updated the code, 
thanks for the review!

Learned a lot :)

TzlilSwimmer123 (changes requested):
Yay almost perfect!
Can you plz remove the go.sum and go.mod changes? we don't use the reflection pkg anumoye

TzlilSwimmer123 (approved):

PR #581, opened 2022-04-20
feat: issue#573 support JUnit output format

### LINKED ISSUE #573: Support JUnit output format
**Motivation**
Push forward a common requested [eco system to support](https://github.com/datreeio/datree/issues/125)

**DOD**
Users are able to set their output format to JUnit by setting the value of -o argument to JUnit

royhadad:


bl/evaluation/printer.go
```diff
@@ -73,8 +74,12 @@ func PrintResults(resultsData *PrintResultsData) error {
 			return jsonOutput(&formattedOutput)
 		} else if resultsData.OutputFormat == "yaml" {
 			return yamlOutput(&formattedOutput)
-		} else {
+		} else if resultsData.OutputFormat == "xml" {
 			return xmlOutput(&formattedOutput)
+		} else if resultsData.OutputFormat == "JUnit" {
+			return jUnitOutput(&formattedOutput)
+		} else {
+			panic(errors.New("invalid output format"))
```
TzlilSwimmer123, line 82: The panic behavior is new,  it's ok by the product team?

royhadad: this should never ever happen (because it is validated way before) - so if it does, we will get notified via pagerduty

TzlilSwimmer123: BTW if you try to run datree test with -o that doesn't support we fail the run: `./datree test internal/fixtures/kube/k8s-demo.yaml -o blump
`:
`Error: Invalid --output option - "blump"
Valid output values are - simple, yaml, json, xml, JUnit`
So maybe this panic is redundant?

Makefile
```diff
@@ -30,3 +30,6 @@ set-token:
 
 publish:
 	go run -tags=staging -ldflags="-X github.com/datreeio/datree/cmd.CliVersion=0.0.1" main.go publish ./internal/fixtures/policyAsCode/valid-schema.yaml
+
+datree_test_to_JUnit_report: # install junit2html first: https://github.com/inorton/junit2html
```
TzlilSwimmer123, line 34: this is nice but each developer who will want to use it will have to download the pkg first which can be confusing w/o looking into this file

cmd/test/main.go
```diff
@@ -401,7 +401,7 @@ type EvaluationResultData struct {
 }
 
 func evaluate(ctx *TestCommandContext, filesPaths []string, prerunData *TestCommandData) (EvaluationResultData, error) {
-	isInteractiveMode := (prerunData.Output != "json") && (prerunData.Output != "yaml") && (prerunData.Output != "xml")
+	isInteractiveMode := (prerunData.Output != "json") && (prerunData.Output != "yaml") && (prerunData.Output != "xml") && (prerunData.Output != "JUnit")
```
TzlilSwimmer123, line 404: Consider extracting `prerunData.Output` to a variable

cmd/test/main.go
```diff
@@ -71,10 +71,10 @@ func (flags *TestCommandFlags) Validate() error {
 	outputValue := flags.Output
 
 	if outputValue != "" {
-		if (outputValue != "simple") && (outputValue != "json") && (outputValue != "yaml") && (outputValue != "xml") {
+		if (outputValue != "simple") && (outputValue != "json") && (outputValue != "yaml") && (outputValue != "xml") && (outputValue != "JUnit") {
```
TzlilSwimmer123, line 74: I am thinking about creating an array of valid outputs and iterate it instead of adding all the time `&& (outputValue != "someValidOutput")`. What do you think? this logic is repeated in serval places in this PR

cmd/test/main.go
```diff
@@ -71,10 +71,10 @@ func (flags *TestCommandFlags) Validate() error {
...
 
 			return fmt.Errorf("Invalid --output option - %q\n"+
-				"Valid output values are - simple, yaml, json, xml\n", outputValue)
+				"Valid output values are - simple, yaml, json, xml, JUnit\n", outputValue)
```
TzlilSwimmer123, line 77: If we will have an array we can print it's values here

bl/evaluation/printer.go
TzlilSwimmer123: That's the reason I feel it's redundant, but your call

royhadad: yep it's totally redundant and should never be reached, if this piece of code gets reached it means we f*** up and PagerDuty will alert us
The reason I still want it here is because the `PrintResults` function doesn't know for sure it will never reach this line of code, we only know it higher up the function hierarchy

cmd/test/main.go
royhadad: ABSOLUTLY!

I promise to do a separate refactor PR after this one, I want to keep the PR lean

cmd/test/main.go
royhadad: will be refactored after this pr

bl/evaluation/junit_output.go
```diff
@@ -0,0 +1,162 @@
...
+// https://llg.cubic.org/docs/junit/
+// https://www.ibm.com/docs/en/developer-for-zos/14.2.0?topic=formats-junit-xml-format
+
+type JUnitOutput struct {
+	XMLName  xml.Name    `xml:"testsuites"`
+	Name     string      `xml:"name,attr"`
+	Tests    int         `xml:"tests,attr"`
+	Failures int         `xml:"failures,attr"`
+	Skipped  int         `xml:"skipped,attr"`
+	Suites   []testSuite `xml:"testsuite"`
```
shmuelsa: `TestSuites`

bl/evaluation/junit_output.go
```diff
@@ -0,0 +1,162 @@
...
+		Properties: &[]property{{
+			Name:  "policyName",
+			Value: formattedOutput.PolicySummary.PolicyName,
+		}, {
+			Name:  "totalRulesInPolicy",
+			Value: strconv.Itoa(formattedOutput.PolicySummary.TotalRulesInPolicy),
+		}, {
+			Name:  "totalRulesFailed",
+			Value: strconv.Itoa(formattedOutput.PolicySummary.TotalRulesFailed),
+		}, {
```
shmuelsa: change places between `totalRulesFailed`  <-> `totalSkippedRules`

shmuelsa: also in our xml output :)

bl/evaluation/printer.go
```diff
@@ -115,7 +120,15 @@ func yamlOutput(formattedOutput *FormattedOutput) error {
...
 func xmlOutput(formattedOutput *FormattedOutput) error {
-	xmlOutput, err := xml.MarshalIndent(formattedOutput, "", "\t")
+	return printAsXml(formattedOutput)
+}
+
+func jUnitOutput(formattedOutput *FormattedOutput) error {
+	return printAsXml(FormattedOutputToJUnitOutput(*formattedOutput))
+}
+
+func printAsXml(output interface{}) error {
```
shmuelsa, line 130: AMAZING!

shmuelsa (dismissed):

shmuelsa (approved):

PR #582, opened 2022-04-21
refactor: output flag code duplication

royhadad:


bl/evaluation/output_flag.go
```diff
@@ -0,0 +1,24 @@
+package evaluation
+
+var FormattedOutputOptions = []string{"yaml", "json", "xml", "JUnit"}
```
shmuelsa, line 3: you done it right.

cmd/test/main_test.go
```diff
@@ -814,15 +814,15 @@ func executeTestCommand(ctx *TestCommandContext, args []string) error {
 
 func test_testCommand_output_flags_validation(t *testing.T, ctx *TestCommandContext) {
 
-	validOutputValues := [5]string{"simple", "json", "yaml", "xml", "JUnit"}
+	validOutputValues := [...]string{"", "simple", "json", "yaml", "xml", "JUnit"}
```
shmuelsa, line 817: [...]


Alt Text

shmuelsa: `"" ` - is for empty case right?

shmuelsa (approved):

cmd/test/main_test.go
royhadad: yep!

PR #585, opened 2022-04-25
feat: save and use default rules from datree dir

shalev007:


cmd/test/main.go
```diff
@@ -515,3 +519,18 @@ func wereViolationsFound(validationManager *ValidationManager, results *evaluati
...
+	}
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return
+	}
+
+	defaultRulesFilePath := filepath.Join(homeDir, ".datree", "defaultRules.yaml")
+
+	ioutil.WriteFile(defaultRulesFilePath, []byte(preRunPolicies), 0644)
```
royhadad: I think we want to add
```
if err != nil {
    // do nothing
}
```

royhadad: what does 0644 mean? can you put it in a constant for readability?

cmd/test/main.go
```diff
@@ -515,3 +519,18 @@ func wereViolationsFound(validationManager *ValidationManager, results *evaluati
 		return false
 	}
 }
+
+func saveDefaultRulesAsFile(ctx *TestCommandContext, preRunPolicies string) {
+	if preRunPolicies == "" {
```
royhadad: don't we always get preRunPolicies? when would it be empty?

cmd/test/main.go
```diff
@@ -515,3 +519,18 @@ func wereViolationsFound(validationManager *ValidationManager, results *evaluati
 		return false
 	}
 }
+
+func saveDefaultRulesAsFile(ctx *TestCommandContext, preRunPolicies string) {
```
royhadad: I think you meant to name it:
`preRunPolicies` => `preRunDefaultRulesYaml`

pkg/policy/policy.go
```diff
@@ -33,7 +36,17 @@ type DefaultRuleDefinition struct {
 }
 
 func GetDefaultRules() (*DefaultRulesDefinitions, error) {
-	defaultRulesDefinitions, err := yamlToStruct(defaultRulesYamlContent)
+	configDefaultRulesYamlContent, err := getDefaultRulesFromFile()
+
+	var defaultRulesDefinitions *DefaultRulesDefinitions
+	if err == nil {
+		defaultRulesDefinitions, err = yamlToStruct(configDefaultRulesYamlContent)
+		fmt.Println("Using default rules from file")
```
royhadad: are those 2 logs here on purpose? will they be removed?

pkg/policy/policy.go
```diff
@@ -33,7 +36,17 @@ type DefaultRuleDefinition struct {
 }
 
 func GetDefaultRules() (*DefaultRulesDefinitions, error) {
-	defaultRulesDefinitions, err := yamlToStruct(defaultRulesYamlContent)
+	configDefaultRulesYamlContent, err := getDefaultRulesFromFile()
+
+	var defaultRulesDefinitions *DefaultRulesDefinitions
+	if err == nil {
+		defaultRulesDefinitions, err = yamlToStruct(configDefaultRulesYamlContent)
```
royhadad: you can just do this in both `if` and `else`, might look cleaner
```return yamlToStruct(configDefaultRulesYamlContent)```

**Review by royhadad — CHANGES_REQUESTED** [review]

File: cmd/test/main.go [hunk 6]
```diff
@@ -515,3 +519,22 @@ func wereViolationsFound(validationManager *ValidationManager, results *evaluati
...
+	}
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return
+	}
+
+	defaultRulesFilePath := filepath.Join(homeDir, ".datree", "defaultRules.yaml")
+
+	writeErr := ioutil.WriteFile(defaultRulesFilePath, []byte(preRunPolicies), 0644)
```
royhadad: what does 0644 mean? can you put it in a constant for readability?

cmd/test/main.go
```diff
@@ -515,3 +519,22 @@ func wereViolationsFound(validationManager *ValidationManager, results *evaluati
...
+
+	if writeErr != nil {
+		fmt.Printf("Failed to write default rules to file %s: %s", defaultRulesFilePath, writeErr)
```
royhadad: what are the consequences of printing an error in this case? what will the user experience? are we sure we want that experience?

cmd/test/main.go
shalev007: I don't know but since the cli-service is a different service and could eventually change, I think we cannot depend on it to always have a value

cmd/test/main.go
shalev007: correct

pkg/policy/policy.go
shalev007: Nice catch, Thanks!

pkg/policy/policy.go
shalev007: Correct

cmd/test/main.go
shalev007: Correct

cmd/test/main.go
shalev007: maybe no

royhadad (approved):

PR #597, opened 2022-05-01
feat: pass cli flags as headers on requests

dimabru:


cmd/test/main.go
```diff
@@ -219,6 +222,19 @@ func New(ctx *TestCommandContext) *cobra.Command {
 	return testCommand
 }
 
+func (flags *TestCommandFlags) ToMapping() map[string]interface{} {
+	val := reflect.Indirect(reflect.ValueOf(flags))
```
myishay, line 226: https://stackoverflow.com/questions/18926303/iterate-through-the-fields-of-a-struct-in-go#:~:text=19,a%20new%20option

dimabru: this is only a static reflection of the object type. there's no way to get the value of a field without using `valueOf` that makes the code look more complicated than it was

myishay (dismissed):

cmd/test/main.go
```diff
@@ -219,6 +222,19 @@ func New(ctx *TestCommandContext) *cobra.Command {
...
+	fieldsAmount := val.Type().NumField()
+	flagsByString := make(map[string]interface{})
+
+	for i := 0; i < fieldsAmount; i++ {
+		field := val.Type().Field(i)
+		flagsByString[field.Name] = val.Field(i).Interface()
+	}
```
dimabru, line 233: Your mama is so stupid, only she can understand this code

myishay (approved):

PR #600, opened 2022-05-02
refactor: change the formatting of all error messages 

### LINKED ISSUE #435: Remove new line and capital letters from errors

We want to make sure our error strings follow a set of guidelines to ensure uniformity and good composability. Quoting Go Code Review Comments:
> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use `fmt.Errorf("something bad")` not `fmt.Errorf("Something bad")`, so that `log.Printf("Reading %s: %v", filename, err)` formats without a spurious capital letter mid-message.
> 

Resource: [https://staticcheck.io/docs/checks/#ST1005](https://staticcheck.io/docs/checks/#ST1005)

**Requirements**
Golang basic level.

HillaShx:
closes #435

Hi team!
I was poking around the codebase and replaced all error strings within `fmt.Errorf` according to the instructions on #435. However,  I also found some that were passed in to `t.Errorf` in the testing suites, and it seemed to me that they are completely separate and shouldn't be changed, but I wasn't sure, so I'd appreciate your input.

image

Thanks!


HillaShx:
I saw the failures, and I'm on it!

dimabru (approved):

HillaShx:
Hi @dimabru! Thank you for approving and merging! :)
I just want to make sure it didn't go unnoticed, did you see my comment regarding `t.Error`? Was I right in assuming it shouldn't be changed?

dimabru:
hi @HillaShx :)
thanks for pointing it out, I missed it.
@noaabarki do you think we should open another task to change `t.Error` as well?

PR #623, opened 2022-05-16
fix: identify timeout network error

dimabru:


bl/validation/k8sValidator.go
```diff
@@ -182,7 +182,7 @@ func (val *K8sValidator) validateResource(filepath string) (bool, []error, *vali
 }
 
 func (val *K8sValidator) isNetworkError(errorString string) bool {
-	return strings.Contains(errorString, "no such host") || strings.Contains(errorString, "connection refused")
+	return strings.Contains(errorString, "no such host") || strings.Contains(errorString, "connection refused") || strings.Contains(errorString, "i/o timeout")
```
TzlilSwimmer123: Suggestion:
networkErrStr = ["no such host","connection refused","i/o timeout"] and iterate over it

TzlilSwimmer123 (commented):
Left 2 suggestions which are the same. WDYT?

bl/validation/k8sValidator.go
dimabru: this will force me to add another function to iterate 
https://stackoverflow.com/questions/15323767/does-go-have-if-x-in-construct-similar-to-python
go doesn't have `arr.includes` logic :)

TzlilSwimmer123: I think that iterating twice is good here. if we iterate the same array.

dimabru: done :)

TzlilSwimmer123 (approved):
Coolz

PR #626, opened 2022-05-16
ci: upgrade `trusty` to `focal` ubuntu version in travis ci

### LINKED ISSUE #625: Use new version of Ubuntu in Travis CI
I was working on PR: #593 and while Github actions was running. I noticed we are using a very old version of Ubuntu as build distribution i.e. `Trusty` https://app.travis-ci.com/github/datreeio/datree/builds/250711087#L9

I checked Travis CI documentation, latest version available is [Focal](https://docs.travis-ci.com/user/reference/focal/). We should use `dist: focal` in code here: https://github.com/datreeio/datree/blob/main/.travis.yml#L5

amustaque97:
Fixes: #625

NOTE: I noticed that in the new version of Ubuntu (Focal), in the [documentation](https://docs.travis-ci.com/user/reference/focal/#python-support) it is not mentioned pyenv is removed or installed but while running CI it was failing so I have to add a command to install pyenv with the default env 3.6.10


myishay:
LGTM!

Will update the branch now and if everything will go smoothly we will merge this on Sunday.
Thanks for your contribution!

myishay (approved):

PR #644, opened 2022-05-25
refactor: yaml validation

### LINKED ISSUE #633: JSON Patch format YAML files - yaml validation error: yaml: unmarshal errors: cannot unmarshal !!seq into map[string]interface {}
**Describe the bug**

JSON Patch yaml files cause the following error:

```
[X]  yaml validation error: yaml: unmarshal errors:
  line 29: cannot unmarshal !!seq into map[string]interface {}
```

**To Reproduce**
```
git clone https://github.com/HariSekhon/Kubernetes-configs k8s
cd k8s
datree test *.jsonpatch.yaml
```

Live example can be seen here in my GitHub Actions run:

https://github.com/HariSekhon/Kubernetes-configs/runs/6490327637?check_suite_focus=true#step:5:15

**Expected behavior**

Expected to handle json patches for k8s, or ignore them.

Latest in GitHub Actions datreeio/action-datree

shmuelsa:


bl/validation/k8sValidator.go
```diff
@@ -139,6 +143,19 @@ func (val *K8sValidator) isK8sFile(fileConfigurations []extractor.Configuration)
 	return true
 }
 
+func ExtractK8sKindAndName(content []byte) (string, string) {
+	jsonValues, err := extractor.ParseJsonToK8sValues(content)
+	if err != nil {
+		return "", ""
```
noaabarki: I think it's better to return error in this case

pkg/extractor/extractor.go
```diff
@@ -69,33 +71,49 @@ type FileConfigurations struct {
...
 	configurations, err := extractYamlConfigurations(content)
-	if err != nil {
+	if err != nil && err != io.EOF {
 		return nil, err
 	} else {
 		return configurations, nil
 	}
 }
 
+func ParseJsonToK8sValues(content []byte) (map[string]interface{}, error) {
```
noaabarki: I think the name is a little odd. It's not necessarily JSON it's basically any byte array and it's unclear what is `k8sValues`.

pkg/extractor/extractor.go
```diff
@@ -69,33 +71,49 @@ type FileConfigurations struct {
 
 func ParseYaml(content string) (*[]Configuration, error) {
 	configurations, err := extractYamlConfigurations(content)
-	if err != nil {
+	if err != nil && err != io.EOF {
```
noaabarki: IMHO, `extractYamlConfigurations` should not return `io.EOF` if it's meaninigless.

pkg/cliClient/cliClient_test.go
```diff
@@ -335,12 +337,26 @@ func readMock(path string) ([]extractor.Configuration, error) {
 	yamlDecoder := yaml.NewDecoder(bytes.NewReader(content))
 
 	for {
-		var doc = map[string]interface{}{}
-		err = yamlDecoder.Decode(&doc)
+		var anyTypeStore interface{}
```
noaabarki: Implementing the function again is a bit weird.  Anyway, I think we can delete `readMock`

pkg/extractor/extractor.go
```diff
@@ -69,33 +71,49 @@ type FileConfigurations struct {
...
+			if err == io.EOF {
+				break
+			} else {
+				return nil, err
+			}
 		}
 
-		if len(doc) > 0 {
-			configurations = append(configurations, doc)
+		yamlByteArray, err := yaml.Marshal(anyTypeStore)
```
noaabarki, line 104: I miss your comments, it made it much more clear to understand why we decode every YAML, marshal it and then parse it to JSON.

noaabarki (changes requested):

noaabarki:
I think naming `configuration []byte ` just makes the code much more complex to understand. To me, `Configuration` has no longer meaning at this point, it is basically `[]bytes`. It's not a blocker but we can talk about it and review the full flow to see how we can improve it

bl/validation/k8sValidator.go
```diff
@@ -13,6 +13,11 @@ import (
 	kubeconformValidator "github.com/yannh/kubeconform/pkg/validator"
 )
 
+type MetadataAndKind struct {
```
myishay: no need for this type

pkg/extractor/extractor.go
```diff
@@ -76,29 +84,72 @@ func ParseYaml(content string) (*[]Configuration, error) {
...
+			return nil, err
+		}
+
+		configurations = append(configurations, extractMetadataNameAndKind(jsonByte))
+	}
+
+	return &configurations, nil
+}
+
+func extractMetadataNameAndKind(content []byte) Configuration {
```
myishay: I think we should rename to something like `extractConfigurationK8sData` since it get's now also Annotations, apiVersion etc.

myishay (approved):

noaabarki (approved):

pkg/extractor/extractor.go
```diff
@@ -83,22 +91,55 @@ func extractYamlConfigurations(content string) (*[]Configuration, error) {
...
+		configurations = append(configurations, extractConfigurationK8sData(jsonByte))
+	}
+
+	return &configurations, nil
+}
+
+func extractConfigurationK8sData(content []byte) Configuration {
+	var configuration Configuration
+	var jsonObject map[string]interface{}
+	configuration.Payload = content
```
shalev007, line 123: is Payload the original content (string of the yaml)?
maybe you can change it to a name that explains better what that is like `rawJson`?

bl/evaluation/evaluator.go
```diff
@@ -385,17 +363,11 @@ func addFailedRule(currentFailedRulesByFiles FailedRulesByFiles, fileName string
...
-
-	annotationsMap, ok := configurationMetadata["annotations"].(map[string]interface{})
-	if !ok {
+	if configuration.MetadataName == "" || configuration.Kind == "" {
 		return nil
 	}
 
-	for annotationKey, annotationValue := range annotationsMap {
+	for annotationKey, annotationValue := range configuration.Annotations {
 		if strings.Contains(annotationKey, SKIP_RULE_PREFIX) {
```
shalev007, line 371: read my mind!!!
king 👑

PR #645, opened 2022-05-26
fix: add JUnit testsuite for files that pass policy check

### LINKED ISSUE #640: JUnit output not working with CircleCI expanded insights
Steps to reproduce:
1. add a "run datree" step to a circleci job as follows (in our case its to verify against argo templates):
```
- run:
    name: run datree
    command: |
      curl https://get.datree.io | /bin/bash
      mkdir -p ~/test-results/datree
      datree test -p Argo templates/*.yaml --verbose -o JUnit > ~/test-results/datree/templates.xml
      datree test -p Argo cron/*.yaml --verbose -o JUnit > ~/test-results/datree/cron.xml
```

2. add "store_test_results" step as follows:
``` 
- store_test_results:
     path: ~/test-results/datree/
```

expected: seeing test results in `tests` in CircleCI
actually: image

this guide from CircleCI was used to initiate this integration: https://circleci.com/blog/how-to-output-junit-tests-through-circleci-2-0-for-expanded-insights/

just as a verification effort here are some screenshots of the pipeline:
image
image

royhadad:


bl/evaluation/printer.go
```diff
@@ -27,7 +25,7 @@ type Printer interface {
 
 type PrintResultsData struct {
 	Results               FormattedResults
-	RulesData             []cliClient.RuleData
+	AdditionalJUnitData   AdditionalJUnitData
```
dimabru, line 28: I think renaming this piece of code to `JUnit` data is a bit counter productive. This file is a printer that is responsible for printing all kinds of outputs and I can easily see this becoming an object containing output specific data for other outputs going forward.

Can we somehow pass this data directly to the `FormattedOutputToJUnitOutput`?

dimabru (approved):

PR #648, opened 2022-05-30
fix: fixes for UTF-8 compatibility

### LINKED ISSUE #646: Support UTF-8 Character Set Only
When running datree as part of our pipeline with Octopus deploy our output prints as follows:

image

the same output in a command window on the machine is as follows:

image

It would appear that the `—` is not a UTF-8 character and nore are characters in the table (for example the quotes are `“` and `”` rather than `"`

It would be nice if the character set on the simple output was UTF-8 to ensure full compatibility

BlythMeister:
* Swap `—` with `-`
* `“` and `”` with `"`
* Only use `Rich` coloured table when not in `simple` output mode

Fixes #646

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ It’s far more effective than manual processes, such as sending an email to a s
 
 ## ⚙️ How it Works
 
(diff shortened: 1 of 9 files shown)


myishay:
Hey @BlythMeister , I reviewed your changes. LGTM!
There were some conflicts with main branch plus a small linting problem that I fixed for you.
I'll wait for the tests to pass and we can merge.
Thanks again!

myishay (approved):

eyarz (approved):

PR #650, opened 2022-05-30
feat: create uninstall script during installation

hadar-co:


install.sh
```diff
@@ -1,5 +1,19 @@
 #!/bin/bash
 
+create_uninstall_script()
+{
+    UNINSTALL_SCRIPT="$HOME/.datree/uninstall.sh"
+    touch $UNINSTALL_SCRIPT && chmod +x $UNINSTALL_SCRIPT
+
+    echo "if [ \"\$(id -u)\" -ne 0 ] ; then" >> $UNINSTALL_SCRIPT
```
dimabru: this is a bit hard to understand. I think that using bash convention for file creation with content is a better way to go here. it will make the code more readable.
here's what I'm talking about:
https://stackoverflow.com/questions/13633638/create-file-with-contents-from-shell-script

dimabru (approved):

PR #651, opened 2022-05-30
fix: jUnit dont panic evaluation if invalid yaml or k8s

TzlilSwimmer123:


royhadad:
does this only stop the panic? or also adds something to the output?

bl/evaluation/junit_output.go
```diff
@@ -211,3 +217,67 @@ func findFileInPolicyValidationResults(fileName string, policyValidationResults
...
+			Name: invalidYamlFile.Path,
+		}
+
+		testCase := testCase{
+			Name:      "invalid yaml file",
+			ClassName: "yaml validation",
+			Skipped:   nil,
+			Failure: &failure{
+				Message: "Invalid yaml file",
+				Content: invalidYamlFile.ValidationErrors[0].Error(),
```
TzlilSwimmer123: On yaml validation failure there is only 1 error

TzlilSwimmer123:
@royhadad it stops the panic and also adds test suit for the validations

royhadad:
I don't see a change in the junit test file (expected output), why are the tests passing?

TzlilSwimmer123:
@royhadad it passed until now and will keep passing because we don't test invalid files in other outputs so it's fine we will not test it on JUnit in particular. It works when testing the code manually

royhadad (approved):

PR #663, opened 2022-06-02
build: add second tag with v for each release

dimabru:


royhadad (approved):
you might need to delete some existing tags before merging this, to make it work

PR #690, opened 2022-06-22
feat: add 5 new probe rules

hadar-co:


pkg/defaultRules/defaultRules.yaml
```diff
@@ -1548,3 +1548,273 @@ rules:
       anyOf:
         - $ref: "#/definitions/metadataNamePattern"
         - $ref: "#/definitions/metadataGenerateNamePattern"
+  - id: 55
+    name: Ensure container probes have an initial delay configured
+    uniqueName: CONTAINERS_INCORRECT_INITIALDELAYSECONDS_VALUE
+    enabledByDefault: true
```
eyarz: the rules should not be enabled by default

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1548,3 +1548,273 @@ rules:
...
+    documentationUrl: "https://hub.datree.io/built-in-rules/ensure-initial-probe-delay"
+    messageOnFailure: "Incorrect value for key `initialDelaySeconds` - set to prevent unwanted default behavior"
```
eyarz: it's not an incorrect value, it's a missing value (applied to all rules)

eyarz: "set to prevent unwanted default behavior" - this is too generic. let's try to make it more informative and customized for each rule.

hadar-co: > it's not an incorrect value, it's a missing value (applied to all rules)

@eyarz 
The rule will fail not only if the value is missing, but also if it is incorrect. So "missing" is a particular case of "incorrect", but not the other way around. A user who has this key configured but with a wrong value might be confused to see "missing value". WDYT?

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1548,3 +1548,273 @@ rules:
       anyOf:
         - $ref: "#/definitions/metadataNamePattern"
         - $ref: "#/definitions/metadataGenerateNamePattern"
+  - id: 55
```
royhadad, line 1551: missing unit tests for new rules, I suggest starting with creating the tests, to assist the development process.

royhadad (changes requested):

pkg/policy/tests/59-fail.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: apps/v1
```
royhadad, line 1: is there any way to make the test files more concise? (minimal file to fail/pass every rule, files should still pass k8s validation), this will make it easier to understand 😊

pkg/defaultRules/defaultRules.yaml
eyarz: good point. I agree.
the best option will be to break this rule into two (one for "exists" and for "correct value"), but I feel it will be too spammy/redundant to the users.

hadar-co: Another option is to change the failure message to:
"Missing or incorrect value for key `initialDelaySeconds`..."

Either that or just "Incorrect" are both good IMO.

pkg/policy/tests/59-fail.yaml
hadar-co: Yeah that makes sense, will do that with the next rules:)

pkg/defaultRules/defaultRules.yaml
eyarz: we don't have any other rule that behaved like that? 
I don't want to make sure we will not create a discrepancy here.

hadar-co: Here are 2 examples using "Incorrect" when the value could also be missing:
https://hub.datree.io/built-in-rules/ensure-configmap-is-recognized-by-argocd
https://hub.datree.io/built-in-rules/ensure-read-only-filesystem

Looks like we're already using this behavior:)

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1548,3 +1548,273 @@ rules:
...
+    name: Ensure container probes have an initial delay configured
+    uniqueName: CONTAINERS_INCORRECT_INITIALDELAYSECONDS_VALUE
+    enabledByDefault: false
+    documentationUrl: "https://hub.datree.io/built-in-rules/ensure-initial-probe-delay"
+    messageOnFailure: "Incorrect value for key `initialDelaySeconds` - set to prevent unwanted default initial delay behavior"
```
eyarz: I think the message is a little confusing.
setting the value to default is not the problem, the problem is that you don't have it set explicitly.

how about that?
Incorrect value for key `initialDelaySeconds` - set explicitly to control the start time before a probe is initiated (min 0)

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1548,3 +1548,273 @@ rules:
       anyOf:
         - $ref: "#/definitions/metadataNamePattern"
         - $ref: "#/definitions/metadataGenerateNamePattern"
+  - id: 55
+    name: Ensure container probes have an initial delay configured
```
eyarz: I hate to be "that guy" but we called it "ensure each container... has..." in all the other rules:
image

eyarz (dismissed):

royhadad (approved):

PR #693, opened 2022-06-23
refactor: unify duplicate logic between `test` & `kustomize`

royhadad:
From now on the code inside the cobra `Run` should be as lean as possible, and shared logic should be added to `TestWrapper` (or deeper).
This also fixes a bug, where `datree kustomize test` did not save the defaultRules.yaml file
If you have a better name, other then `TestWrapper` - please let me know 😊


myishay (approved):

myishay:
Approved. Can you please make sure the kustomize flow didn't break?

PR #734, opened 2022-07-20
refactor: remove unneeded accountExists from /prerun response

royhadad:


dimabru (approved):

royhadad:
we might want it in the future

PR #756, opened 2022-08-09
fix: handle http response code 4XX/5XX as errors

### LINKED ISSUE #746: Offline mode does not work.
Attempting to run the tool in an airgapped environment results in an error.

```
tech@WSAMZN-7N8TKGIS:/mnt/d/Workspace/tech/schemas$ kubectl-datree test -s 1.22.6 --verbose --no-record --schema-location ./kubernetes-json-schema/ -- -n nginx-ingress
Fetching resources, this may take some time depending on the amount of resources in your cluster...

invalid character '<' looking for beginning of value
```

I believe this is attempting to reach out to the internet and is getting a 403 'forbidden' html page response from the local in-line proxy as it can't reach out to the internet. That's where the invalid character '<' comes from, an html tag.

myishay:


TzlilSwimmer123 (approved):
Dibur

PR #760, opened 2022-08-16
feat: write json output to local file on every policy check

hadar-co:


hadar-co:
@myishay 
We shouldn't exit if one of the steps fails (since this will not display any output to the user).

pkg/evaluation/printer.go
```diff
@@ -54,7 +55,37 @@ type textOutputData struct {
...
+	if err != nil {
+		return
+	}
+
+	jsonOutput, err := GetjsonResult(resultsData)
+	if err != nil {
+		return
+	}
+
+	lastPolicyCheckPath := homeDir + "/.datree/lastPolicyCheck.json"
```
myishay, line 69: might not work with windows. should be checked

myishay (approved):

PR #769, opened 2022-08-27
feat: add "ensure seccomp not unconfined" rule

Heydad-Helfer:
Added the new schema, used defs for reusability. 
Documentation link is still needed.

### DIFF
--- a/pkg/defaultRules/defaultRules.yaml
+++ b/pkg/defaultRules/defaultRules.yaml
@@ -1870,11 +1870,11 @@ rules:
       items:
         $ref: "#"
   - id: 61
-    name: "Prevent setting 'Unconfined' as seccomp profile"
+    name: "Prevent containers from having unnecessary system call privileges"
     uniqueName: CONTAINERS_INCORRECT_SECCOMP_PROFILE
     enabledByDefault: false
-    documentationUrl: "https://www.datree.io/resources/eks-best-practices-you-should-know"
-    messageOnFailure: "seccompProfile was omitted or was explicitly set to 'Unconfined' - set an explicit value to increase the pod or container's isolation"
+    documentationUrl: "https://hub.datree.io/built-in-rules/prevent-system-call-privileges"
+    messageOnFailure: "Incorrect value for key seccompProfile - set an explicit value to prevent malicious use of system calls within the container"
     category: Containers
     schema:
       definitions:
--- a/pkg/defaultRules/defaultRules.yaml
+++ b/pkg/defaultRules/defaultRules.yaml
@@ -1872,7 +1872,7 @@ rules:
   - id: 61
     name: Ensure seccomp profile must not be explicitly set to Unconfined
     uniqueName: CONTAINERS_INCORRECT_SECCOMP_PROFILE
-    enabledByDefault: true
+    enabledByDefault: false
     documentationUrl: "https://www.datree.io/resources/eks-best-practices-you-should-know"
     messageOnFailure: "Prevent setting 'Unconfined' as seccomp profile - unset it to retain the pod's or container's isolation"
     category: Containers
--- a/pkg/defaultRules/defaultRules.yaml
+++ b/pkg/defaultRules/defaultRules.yaml
@@ -1879,25 +1879,29 @@ rules:
     schema:
       definitions:
         seccompPatternInSpec:
-          properties:
-            spec:
-              $ref: "#/$defs/securityContextSeccomp"
+          if: *standardKinds
+          then:
+            properties:
+              spec:
+                $ref: "#/$defs/securityContextSeccomp"
         seccompPatternInContainer:
-          properties:
-            spec:
-              properties:
-                containers:
-                  type: array
-                  items:
-                    $ref: "#/$defs/securityContextSeccomp"
-                initContainers:
-                  type: array
-                  items:
-                    $ref: "#/$defs/securityContextSeccomp"
-                ephemeralContainers:
-                  type: array
-                  items:
-                    $ref: "#/$defs/securityContextSeccomp"
+          if: *standardKinds
+          then:
+            properties:
+              spec:
+                properties:
+                  containers:
+                    type: array
+                    items:
+                      $ref: "#/$defs/securityContextSeccomp"
+                  initContainers:
+                    type: array
+                    items:
+                      $ref: "#/$defs/securityContextSeccomp"
+                  ephemeralContainers:
+                    type: array
+                    items:
+                      $ref: "#/$defs/securityContextSeccomp"
 
       allOf:
         - $ref: "#/definitions/seccompPatternInSpec"
--- a/pkg/defaultRules/defaultRules.yaml
+++ b/pkg/defaultRules/defaultRules.yaml
@@ -1964,6 +1964,3 @@ rules:
                         enum:
                           - "unconfined"
                           - "Unconfined"
-
-
-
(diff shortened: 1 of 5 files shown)


eyarz:
@Heydad-Helfer Thank you for the PR!
Do you also plan to write the docs for this rule?
https://github.com/datreeio/docs

Heydad-Helfer:
> @Heydad-Helfer Thank you for the PR! Do you also plan to write the docs for this rule? https://github.com/datreeio/docs

Hi. I can add this, but it'll take me some time. Do you want to wait on the merge until a documentation is available?

eyarz:
NP, we can wait :)
it will take us a few days to review your PR anyway...

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1869,3 +1869,58 @@ rules:
         $ref: "#"
       items:
         $ref: "#"
+  - id: 61
+    name: Ensure seccomp profile must not be explicitly set to Unconfined
```
eyarz: I think it will be more clear to the user if you will use "prevent" instead of "ensure ... not".
so maybe "Prevent setting 'Unconfined' as seccomp profile"?

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1869,3 +1869,58 @@ rules:
...
+    uniqueName: CONTAINERS_INCORRECT_SECCOMP_PROFILE
+    enabledByDefault: true
+    documentationUrl: "https://www.datree.io/resources/eks-best-practices-you-should-know"
+    messageOnFailure: "Seccomp profile is explicitly set to Unconfined - unset it to retain the pod's or container's isolation"
+    category: Containers
+    schema:
```
eyarz, line 1879: from the K8s docs:
`...you can configure Kubernetes to use the profile that the container runtime prefers by default, rather than falling back to Unconfined`

so, basically, not setting `seccompProfile` means you are using `Unconfined` as your seccomp profile.
also, `seccompProfile` supports only the values "Localhost", "RuntimeDefault", "Unconfined", so maybe a validation that "Localhost" or "RuntimeDefault" value is set will be more comprehensive.
WDYT?

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1869,3 +1869,58 @@ rules:
         $ref: "#"
       items:
         $ref: "#"
+  - id: 61
+    name: Ensure seccomp profile must not be explicitly set to Unconfined
+    uniqueName: CONTAINERS_INCORRECT_SECCOMP_PROFILE
+    enabledByDefault: true
```
eyarz: let's set it to false because it's too aggressive rule to enforce on all users

eyarz (changes requested):
@Heydad-Helfer please review my comments :)

pkg/defaultRules/defaultRules.yaml
Heydad-Helfer: So should check be that the seccompProfile *does* defined in the config, and that it's only "Localhost" or "RuntimeDefault"?

pkg/defaultRules/defaultRules.yaml
Heydad-Helfer: Modified the message

pkg/defaultRules/defaultRules.yaml
Heydad-Helfer: Set the enableOnDefault to false

pkg/defaultRules/defaultRules.yaml
eyarz: I think this should be the logic:
seccompProfile has to be set && it has to be "Localhost" or "RuntimeDefault

Heydad-Helfer: OK, so the policy should check that seccompProfile is either defined globally and container overrides are optional, or is must be defined for each and every container. And in any case, it must not be "Unconfined".

I'll start working on it.

eyarz:
@Heydad-Helfer is this ready for code review?

Heydad-Helfer:
Hi, unfortunately not yet, as I didn't have much time to finish this. I'll
work on it this weekend.

pkg/defaultRules/defaultRules.yaml
Heydad-Helfer: @eyarz please check the policy

eyarz:
@Heydad-Helfer no worries, take your time.
just ping me when I should review it again :)

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1869,3 +1869,96 @@ rules:
         $ref: "#"
       items:
         $ref: "#"
+  - id: 61
+    name: "Prevent setting 'Unconfined' as seccomp profile"
```
hadar-co: The name should be about what this rule does, and not how it does it. Something like:
"Prevent containers from having unnecessary system call privileges"
what do you think?

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1869,3 +1869,96 @@ rules:
...
+    uniqueName: CONTAINERS_INCORRECT_SECCOMP_PROFILE
+    enabledByDefault: false
+    documentationUrl: "https://www.datree.io/resources/eks-best-practices-you-should-know"
+    messageOnFailure: "seccompProfile was omitted or was explicitly set to 'Unconfined' - set an explicit value to increase the pod or container's isolation"
```
hadar-co: Also, to keep a consistent format of our rule docs, I think this should be:
"Incorrect value for key `seccompProfile` - set an explicit value to prevent malicious use of system calls within the container"

hadar-co:
@Heydad-Helfer Hey Heydad, left you some comments on your PR:)

Also, we created the docs for this rule:
https://github.com/datreeio/docs/pull/89

Heydad-Helfer:
Hi. Thanks.
I'll make the changes.

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1869,3 +1869,96 @@ rules:
...
+                  ephemeralContainers:
+                    type: array
+                    items:
+                      $ref: "#/$defs/securityContextSeccomp"
+
+      allOf:
+        - $ref: "#/definitions/seccompExplicit"
+        - $ref: "#/definitions/seccompPatternInSpec"
+        - $ref: "#/definitions/seccompPatternInContainer"
+
```
hadar-co, line 1934: Add the following rows (otherwise this will not work for Deployments):
additionalProperties:
  $ref: "#"
items:
  $ref: "#"

hadar-co (changes requested):

hadar-co:
@Heydad-Helfer Thanks, just a couple of last things - 
1. I left a comment regarding some lines to add to support Deployments as well (currently the rule will only work for Pods).
2. All of our rules have test files under pkg/policy/tests (1 passing manifest and 1 failing for each rule). Can you please add test files for your rule? They should be called 61-pass.yaml and 61-fail.yaml. 

And that should be it! 🥳

Heydad-Helfer:
Hi @hadar-co , I pushed the fixes. Please review.
Thanks!

hadar-co (approved):

eyarz (approved):

hadar-co:
@Heydad-Helfer Awesome:) thanks for the PR 🤘🏽

PR #784, opened 2022-09-11
fix: same $HOME between install and uninstall

### LINKED ISSUE #783: uninstall script fails to clean .datree files when sudo doesn't preserve $HOME
**Describe the bug**
The directory and contents at `$HOME/.datree` fail to be deleted when running uninstall script. 

**Investigation:**
The created uninstall script calls `$HOME` [**when the script is run**](https://github.com/datreeio/datree/blob/7e10999830da1f4379fb08befe6051ae70b4e73b/install.sh#L13) at uninstall time. 
This `$HOME` may be different to the `$HOME` when running the install script, for instance when both of these conditions are met:
* using `sudo` when running the one and not the other [[It should be mentioned that the script forces the user to run as root](https://github.com/datreeio/datree/blob/7e10999830da1f4379fb08befe6051ae70b4e73b/install.sh#L10)]. 

AND:
* using a version of sudo that doesn't preserve $HOME (see: https://askubuntu.com/questions/1186999/how-does-sudo-handle-home-differently-since-19-10).

**To Reproduce**
1. Run install script without sudo.
2. Run uninstall script with version of sudo that doesn't preserve `$HOME`

**Expected behavior**
The uninstall script should delete the `.datree` directory, regardless of the run-time value(s) of $HOME.

**Desktop:**
 - OS: WSL (Ubuntu 20.04)

**Datree version:**
 - Version: 1.6.19

netanel-haber:
fixes issue https://github.com/datreeio/datree/issues/783.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-DATREE_HOME="$HOME/.datree"
-
 create_uninstall_script()
 {
     UNINSTALL_SCRIPT="$HOME/.datree/uninstall.sh"
@@ -55,7 +53,7 @@ fi
 
 unzip -qq $OUTPUT_BASENAME_WITH_POSTFIX -d $OUTPUT_BASENAME
 
-mkdir -p $DATREE_HOME
+mkdir -p ~/.datree
 
 rm -f /usr/local/bin/datree 2> /dev/null || sudo rm -f /usr/local/bin/datree
 cp $OUTPUT_BASENAME/datree /usr/local/bin 2> /dev/null || sudo cp $OUTPUT_BASENAME/datree /usr/local/bin
@@ -64,7 +62,7 @@ rm $OUTPUT_BASENAME_WITH_POSTFIX
 rm -rf $OUTPUT_BASENAME
 
 # download and save demo file
-curl -s https://get.datree.io/k8s-demo.yaml > $DATREE_HOME/k8s-demo.yaml
+curl -s https://get.datree.io/k8s-demo.yaml > ~/.datree/k8s-demo.yaml
 
 # create uninstall script
 create_uninstall_script
--- a/install.sh
+++ b/install.sh
@@ -73,7 +73,7 @@ echo -e "\033[32m[V] Finished Installation\033[0m"
 
 echo
 
-echo -e "\033[35m Usage: $ datree test $DATREE_HOME/k8s-demo.yaml \033[0m"
+echo -e "\033[35m Usage: $ datree test ~/.datree/k8s-demo.yaml \033[0m"
 
 echo -e "\033[35m Using Helm? => https://github.com/datreeio/helm-datree \033[0m"
 
--- a/install.sh
+++ b/install.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
 
+DATREE_HOME="$HOME/.datree"
+
 create_uninstall_script()
 {
-    UNINSTALL_SCRIPT="$HOME/.datree/uninstall.sh"
+    UNINSTALL_SCRIPT="$DATREE_HOME/uninstall.sh"
     touch $UNINSTALL_SCRIPT && chmod +x $UNINSTALL_SCRIPT
 
     cat >> $UNINSTALL_SCRIPT << 'END'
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,18 +1,20 @@
...
-    UNINSTALL_SCRIPT="$HOME/.datree/uninstall.sh"
+    UNINSTALL_SCRIPT="$DATREE_HOME/uninstall.sh"
     touch $UNINSTALL_SCRIPT && chmod +x $UNINSTALL_SCRIPT
 
     cat >> $UNINSTALL_SCRIPT << 'END'
     if [ "$(id -u)" -ne 0 ] ; then
     echo "This script must be executed with root privileges." && exit 1
     fi
     rm -f /usr/local/bin/datree
-    rm -rf $HOME/.datree
```
hadar-co: Why move these 2 lines out of the `cat` section? You can leave them in as:
```
rm -rf "$DATREE_HOME"
echo "Datree was successfully uninstalled."
```

install.sh
```diff
@@ -71,7 +73,7 @@ echo -e "\033[32m[V] Finished Installation\033[0m"
 
 echo
 
-echo -e "\033[35m Usage: $ datree test ~/.datree/k8s-demo.yaml \033[0m"
+echo -e "\033[35m Usage: $ datree test $DATREE_HOME/k8s-demo.yaml \033[0m"
```
hadar-co: This is output for the user, should stay as `~` so it's shorter

hadar-co:
hey @netanel-haber! thanks for the PR, left some comments:)

netanel-haber:
> hey @netanel-haber! thanks for the PR, left some comments:)

I don't see any comments?

hadar-co (changes requested):

hadar-co:
> > hey @netanel-haber! thanks for the PR, left some comments:)
> 
> I don't see any comments?

My bad, try now:)

install.sh
netanel-haber: Thats exactly what this PR fixes - the tricky thing with the `cat >> X.sh << 'END' ... END` trick is that it outputs **the literal text** into the file, without interpolation. My fix ensures that the initial install.sh $HOME is captured, and input into the uninstall.sh file.

[This video](https://user-images.githubusercontent.com/58652339/189615325-400cd1f9-c350-4c8c-98fb-0a8c8bb0d63c.mov) shows that $HOME isn't evaluated when catting the string, rather `$HOME` is input **literally** into the file, which is the exact problem in the first place.

So in your suggestion, the string `$DATREE_HOME` will be input into the file - and this variable obviously won't be available when the script is run.

hadar-co: Yup, I see what you mean. Thanks for the detailed response👍🏼

hadar-co (dismissed):

install.sh
hadar-co: @netanel-haber 
After some reading&testing, it seems that the fix is much simpler. The `cat >> X.sh << END ... END` (AKA [here-document](https://tldp.org/LDP/abs/html/here-docs.html)) does interpolate variables by default. To disable the interpolation, you add quotes around the `END`, which is what we originally had in the script, so we should simply remove them.
If you want to prevent a single instance from being interpolated - like in our case the `$(id -u)`, you can simply escape it using `\`. Put all of this together and the fix should look like this:
```
UNINSTALL_SCRIPT="$HOME/.datree/uninstall.sh"
    touch $UNINSTALL_SCRIPT && chmod +x $UNINSTALL_SCRIPT

    cat > $UNINSTALL_SCRIPT << END
    if [ "\$(id -u)" -ne 0 ] ; then
    echo "This script must be executed with root privileges." && exit 1
    fi
    rm -f /usr/local/bin/datree
    rm -rf $HOME/.datree
    echo "Datree was successfully uninstalled."
END
``` 

Bash is tricky 😈

@myishay FYI

myishay: @hadar-co Interesting!

@netanel-haber WDYT?

netanel-haber: Much nicer. Learned something too.

myishay: @netanel-haber So would you like to make the desired changes? I think we can merge after that 🚀

netanel-haber: Sure

hadar-co (approved):

PR #798, opened 2022-09-15
fix: add --skip-validation flag

myishay:


cmd/test/main.go
```diff
@@ -70,9 +71,19 @@ func NewTestCommandFlags() *TestCommandFlags {
...
 
+func validateSkipValidationFlag(flags *TestCommandFlags) error {
+	if flags.SkipValidation != "" && flags.SkipValidation != "schema" {
+		return fmt.Errorf("invalid --skip-validation option - %q\n"+
+			"Valid values are - schema", flags.SkipValidation)
+
+	}
+	return nil
+}
+
```
TzlilSwimmer123: Maybe this will be more robust for the future:
```
supportedSkipValidation := []string{"schema"}
	
if slices.Contains(supportedSkipValidation, flags.SkipValidation) {
return fmt.Errorf("invalid --skip-validation option - %q\n"+
	"Valid values are - schema", flags.SkipValidation)

}
```
When you print the valid values you can use the string array

cmd/test/main.go
```diff
@@ -292,6 +311,7 @@ func GenerateTestCommandData(testCommandFlags *TestCommandFlags, localConfigCont
 		ClientId:              localConfigContent.ClientId,
 		RegistrationURL:       evaluationPrerunDataResp.RegistrationURL,
 		PromptRegistrationURL: evaluationPrerunDataResp.PromptRegistrationURL,
+		SkipSchemaValidation:  testCommandFlags.SkipValidation == "schema",
```
TzlilSwimmer123, line 317: what will happen in the future if we will have an array of possible validations?

cmd/test/main.go
TzlilSwimmer123: We can extract all the validation to a sort of a util, can open an OSS issue for that

cmd/test/main.go
```diff
@@ -484,7 +509,7 @@ func evaluate(ctx *TestCommandContext, filesPaths []string, testCommandData *Tes
 	policyName := testCommandData.Policy.Name
 
 	policyCheckData := evaluation.PolicyCheckData{
-		FilesConfigurations: validationManager.ValidK8sFilesConfigurations(),
+		FilesConfigurations: validationManager.K8sFilesConfigurationsForPolicyCheck(),
```
TzlilSwimmer123: Why the `PolicyCheck` is important here?

TzlilSwimmer123 (commented):
left some questions

cmd/test/main.go
myishay: In this case we will change the flags internally, but it won't damage the user experience since the real flag will stay `--skip-validation=schema`

cmd/test/main.go
myishay: yeah it might not be the best name. I want this name to represent the files configurations that are valid or skipped and will be evaluated by the datree policy check. do you have a better name for that?

shmu3l (approved):

PR #804, opened 2022-09-25
feat: add Strict policy

shalev007:


pkg/defaultPolicies/defaultPolicies.yaml
```diff
@@ -19,7 +19,7 @@ policies:
       - identifier: SERVICE_INCORRECT_TYPE_VALUE_NODEPORT
         messageOnFailure: Incorrect value for key `type` - `NodePort` will open a port on all nodes where it can be reached by the network external to the cluster
       - identifier: CRONJOB_INVALID_SCHEDULE_VALUE
-        messageOnFailure: 'Incorrect value for key `schedule` - the (cron) schedule expressions is not valid and, therefore, will not work as expected'
+        messageOnFailure: "Incorrect value for key `schedule` - the (cron) schedule expressions is not valid and, therefore, will not work as expected"
```
TzlilSwimmer123: Why do we need "" here and in there other no?

TzlilSwimmer123 (changes requested):
just making sure we need "" or not in each `messageOnFailure`

shmu3l (approved):

TzlilSwimmer123 (approved):

PR #809, opened 2022-10-13
feat: support upgrade command

### LINKED ISSUE #752: Support datree upgrade command

Add new `upgrade` command to install the latest datree version. 

- The command should support only macOS & Linux.
- The command should check if the user installed Datree with brew and if so, the command should fail and return a message “Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree”
- The command should not support pinned or user-defined versions.

**Requirements**
Golang mid-level.

**Reference:** [https://github.com/cli/cli/blob/trunk/pkg/cmd/extension/command.go](https://github.com/cli/cli/blob/trunk/pkg/cmd/extension/command.go)

amustaque97:
closes #752 

Here is the quick demo screencast — `datree upgrade` command

https://user-images.githubusercontent.com/21127788/195978978-ccd97fff-969f-45db-94b0-28b992df238f.mov

### DIFF
--- a/cmd/upgrade/main.go
+++ b/cmd/upgrade/main.go
@@ -1,0 +1,48 @@
+package upgrade
+
+import (
+	"fmt"
+
+	"github.com/datreeio/datree/pkg/cliClient"
+	"github.com/datreeio/datree/pkg/upgradeManager"
+	"github.com/spf13/cobra"
+)
+
+type Printer interface {
+	PrintMessage(messageText string, messageColor string)
+}
+type UpgradeCommandContext struct {
+	CliVersion       string
+	Printer          Printer
+	UpgradeCliClient *cliClient.CliClient
+}
+
+func upgrade(ctx *UpgradeCommandContext) {
+	fmt.Println(ctx.CliVersion)
+}
+
+func New(ctx *UpgradeCommandContext) *cobra.Command {
+	m := upgrademanager.NewUpgradeManager()
+	var upgradeCmd = &cobra.Command{
+		Use:   "upgrade",
+		Short: "Upgrade datree to the latest version",
+		Long:  "Upgrade datree to the latest version",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) > 0 {
+				ctx.Printer.PrintMessage("Run upgrade command without any argument\n", "error")
+			}
+
+			if m.CheckIfDatreeInstalledUsingBrew() {
+				ctx.Printer.PrintMessage("Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree\n", "error")
+			}
+
+			err := m.Upgrade()
+			if err != nil {
+				err_msg := fmt.Sprintf("Failed to upgrade datree to the latest version, error encountered %s", err)
+				ctx.Printer.PrintMessage(err_msg, "error")
+			}
+
+		},
+	}
+	return upgradeCmd
+}
--- a/cmd/upgrade/main.go
+++ b/cmd/upgrade/main.go
@@ -24,8 +24,8 @@ func New(ctx *UpgradeCommandContext) *cobra.Command {
 		Short: "Upgrade datree to the latest version",
 		Long:  "Upgrade datree to the latest version",
 		Run: func(cmd *cobra.Command, args []string) {
-			if len(args) > 0 {
-				ctx.Printer.PrintMessage("Run upgrade command without any argument\n", "error")
+			if m.CheckIfDatreeInstalledUsingBrew() {
+				ctx.Printer.PrintMessage("Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree\n", "error")
 			} else {
 				if m.CheckIfDatreeInstalledUsingBrew() {
 					ctx.Printer.PrintMessage("Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree\n", "error")
@@ -39,7 +39,6 @@ func New(ctx *UpgradeCommandContext) *cobra.Command {
 					ctx.Printer.PrintMessage("Datree upgraded successfully\n", "green")
 				}
 			}
-
 		},
 	}
 	return upgradeCmd
--- a/cmd/upgrade/main.go
+++ b/cmd/upgrade/main.go
@@ -36,7 +36,7 @@ func New(ctx *UpgradeCommandContext) *cobra.Command {
 						ctx.Printer.PrintMessage(errMsg, "error")
 						return
 					}
-					ctx.Printer.PrintMessage("Datree upgraded successfully\n", "green")
+					ctx.Printer.PrintMessage("[V] Datree upgraded successfully\n", "green")
 				}
 			}
 		},
--- a/cmd/upgrade/main.go
+++ b/cmd/upgrade/main.go
@@ -26,16 +26,16 @@ func New(ctx *UpgradeCommandContext) *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			if len(args) > 0 {
 				ctx.Printer.PrintMessage("Run upgrade command without any argument\n", "error")
-			}
-
-			if m.CheckIfDatreeInstalledUsingBrew() {
-				ctx.Printer.PrintMessage("Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree\n", "error")
-			}
-
-			err := m.Upgrade()
-			if err != nil {
-				err_msg := fmt.Sprintf("Failed to upgrade datree to the latest version, error encountered %s", err)
-				ctx.Printer.PrintMessage(err_msg, "error")
+			} else {
+				if m.CheckIfDatreeInstalledUsingBrew() {
+					ctx.Printer.PrintMessage("Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree\n", "error")
+				} else {
+					err := m.Upgrade()
+					if err != nil {
+						err_msg := fmt.Sprintf("Failed to upgrade datree to the latest version, error encountered %s", err)
+						ctx.Printer.PrintMessage(err_msg, "error")
+					}
+				}
 			}
 
 		},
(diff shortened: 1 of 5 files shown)


shmu3l:
@amustaque97 It looks great! I'll check it out and promote it further in the next few days

shmu3l:
Hey @amustaque97!

i reviewed your PR, regarding the logic for upgrading, we already have it in [`install.sh`](https://github.com/datreeio/datree/blob/main/install.sh), so no need to write it again, you can use the existing script.

WDYT?

amustaque97:
> Hey @amustaque97!
> 
> i reviewed your PR, regarding the logic for upgrading, we already have it in [`install.sh`](https://github.com/datreeio/datree/blob/main/install.sh), so no need to write it again, you can use the existing script.
> 
> WDYT?

I would say, if we have more plans/ideas then we use golang code only because we can extend this functionality more easily

else

bash code is also fine. 


It boils down to personal preference because I personally prefer go code rather than a bash script 😛  because of silly mistakes and hard time in debugging. I'm fine either way. Just that it should solve our purpose.

Looking forward to hearing from you.

myishay:
I think the main concern here is that we don't want to maintain two pieces of code with basically the same logic.
The bash script already exists and has to stay for the initial installation, so even if it might be easier to debug and read/write golang than bash script, it will make us maintain the same logic in 2 different code blocks.

@amustaque97 @shmu3l WDYT?

amustaque97:
Hey @shmu3l and @myishay I have made the code changes and now I'm relying on the `install.sh` script content. I tested it in my local. It solves the purpose. Please review my PR.

cmd/upgrade/main.go
```diff
@@ -0,0 +1,46 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) > 0 {
+				ctx.Printer.PrintMessage("Run upgrade command without any argument\n", "error")
+			} else {
+				if m.CheckIfDatreeInstalledUsingBrew() {
+					ctx.Printer.PrintMessage("Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree\n", "error")
+				} else {
+					err := m.Upgrade()
+					if err != nil {
+						err_msg := fmt.Sprintf("Failed to upgrade datree to the latest version, error encountered %s\n", err)
```
shalev007: This is small, but let's try to keep the variable naming to camel case.
```suggestion
						errMsg := fmt.Sprintf("Failed to upgrade datree to the latest version, error encountered %s\n", err)
```

shalev007 (changes requested):
Looks great!
left a small comment, I'll approve after this minor fix

pkg/upgradeManager/upgradeManager.go
```diff
@@ -0,0 +1,61 @@
...
+	_, err := m.newCommand("brew", "list", "datree").CombinedOutput()
+	return err == nil
+}
+
+func (m *UpgradeManager) Upgrade() error {
+
+	client := m.client
+
+	// fetch the content of the shell script i.e. `install.sh`
+	response, err := client.Request(http.MethodGet, "/datreeio/datree/main/install.sh", nil, map[string]string{})
```
royhadad: This is not bad, however, we should fetch the script from "https://get.datree.io", since this is the actual production install script

pkg/upgradeManager/upgradeManager.go
```diff
@@ -0,0 +1,61 @@
...
+		client:     httpClient.NewClient("https://raw.githubusercontent.com", nil),
+	}
+}
+
+func (m *UpgradeManager) CheckIfDatreeInstalledUsingBrew() bool {
+	_, err := m.newCommand("brew", "list", "datree").CombinedOutput()
+	return err == nil
+}
+
+func (m *UpgradeManager) Upgrade() error {
```
royhadad: This is the installation one-liner from the datree docs:
https://hub.datree.io/#quickstart

This seems a bit cumbersome to me, we can just execute the one-liner using Go's standard library, maybe something like this? What do you think?

```
out, err := exec.Command("bash", "curl https://get.datree.io | /bin/bash").Output()
```

cmd/upgrade/main.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func New(ctx *UpgradeCommandContext) *cobra.Command {
+	m := upgrademanager.NewUpgradeManager()
+	var upgradeCmd = &cobra.Command{
+		Use:   "upgrade",
+		Short: "Upgrade datree to the latest version",
+		Long:  "Upgrade datree to the latest version",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) > 0 {
```
royhadad: This should be checked in the "args" function, check out this example:
https://github.com/datreeio/datree/blob/4cb0e38c252a102ec8bb0b6010ebe2633dc0a19c/cmd/test/main.go#L200

Actually, even better, this check can be deleted completely! Cobra checks it by default

Screen Shot 2022-10-23 at 22 17 28

cmd/upgrade/main.go
```diff
@@ -0,0 +1,46 @@
...
+	PrintMessage(messageText string, messageColor string)
+}
+type UpgradeCommandContext struct {
+	CliVersion       string
+	Printer          Printer
+	UpgradeCliClient *cliClient.CliClient
+}
+
+func New(ctx *UpgradeCommandContext) *cobra.Command {
+	m := upgrademanager.NewUpgradeManager()
```
royhadad, line 21: Because this is an I/O dependency, adding it to the `UpgradeCommandContext` and injecting it from the outside in cmd/root.go makes more sense to me.
It will allow for easier testing

cmd/upgrade/main.go
```diff
@@ -0,0 +1,46 @@
...
+				if m.CheckIfDatreeInstalledUsingBrew() {
+					ctx.Printer.PrintMessage("Looks like you installed Datree via brew, to upgrade datree run - brew upgrade datree\n", "error")
+				} else {
+					err := m.Upgrade()
+					if err != nil {
+						errMsg := fmt.Sprintf("Failed to upgrade datree to the latest version, error encountered %s\n", err)
+						ctx.Printer.PrintMessage(errMsg, "error")
+						return
+					}
+					ctx.Printer.PrintMessage("Datree upgrade successfully\n", "green")
```
royhadad:
```suggestion
					ctx.Printer.PrintMessage("Datree upgraded successfully\n", "green")
```

royhadad (changes requested):

pkg/upgradeManager/upgradeManager.go
amustaque97: Thanks for the heads-up. I think "https://get.datree.io" is short and easy to implement.

cmd/upgrade/main.go
amustaque97: Thank you for the comment. I didn't know that 😅 . I removed the condition.

pkg/upgradeManager/upgradeManager.go
amustaque97: thanks for the review comment. After reading the issue description I thought we want to replicate a similar coding style but now I understood we just want implementation. I got confused 😕 

I have made the required changes. Code is much simpler and short.

amustaque97:
@royhadad, thank you for the brief PR review. I have made the required changes. New code changes work fine, tested locally. Sharing a screenshot for reference. (In the screenshot I uninstalled datree intentionally to show downloading script works fine and similarly, it will upgrade existing datree using shell script)

Screenshot 2022-10-24 at 1 40 30 AM

cmd/upgrade/main.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func New(ctx *UpgradeCommandContext) *cobra.Command {
+	m := upgrademanager.NewUpgradeManager()
+	var upgradeCmd = &cobra.Command{
+		Use:   "upgrade",
+		Short: "Upgrade datree to the latest version",
+		Long:  "Upgrade datree to the latest version",
+		Run: func(cmd *cobra.Command, args []string) {
+			if m.CheckIfDatreeInstalledUsingBrew() {
```
royhadad, line 27: seems like this is a duplicated if statement :)

pkg/upgradeManager/upgradeManager.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (m *UpgradeManager) CheckIfDatreeInstalledUsingBrew() bool {
+	_, err := exec.Command("brew", "list", "datree").Output()
+	return err == nil
+}
+
+func (m *UpgradeManager) Upgrade() error {
+
+	shellScript := exec.Command("curl", "https://get.datree.io")
```
royhadad: This seems to work as well, replacing the entire function with 3 lines of code.

```
oneLineInstallationCommand := exec.Command("bash", "-c", "curl https://get.datree.io | /bin/bash")
oneLineInstallationCommand.Stdout = os.Stdout
return oneLineInstallationCommand.Run()
```

Is there any benefit to using the current 20 lines of code instead of the suggested 3? I'm asking out of curiosity

royhadad (changes requested):
last 2 comments :)

cmd/upgrade/main.go
amustaque97: I don't think so, In the issue, it is mentioned that we need to check if datree is installed using brew or not. In `install.sh` I don't see that logic with the brew command. Correct me If I'm wrong

pkg/upgradeManager/upgradeManager.go
amustaque97: I have barely seen `-c` option and if you're confident about the one-liner. I would love to see it happening. I have changed the code with the given code block. 👍🏻

shalev007 (dismissed):

cmd/upgrade/main.go
royhadad: look at line 27 and line 30, the same `if` statement is written twice :)

royhadad (changes requested):

cmd/upgrade/main.go
amustaque97: oops my mistake, sorry @royhadad, I didn't notice that I messed it up somehow when I took a pull from the remote branch when I committed the suggested changes from your on Github.

I hope everything is fine now.

royhadad:
@amustaque97 since you got reviews from several people already, I didn't want to bother you for to 4th time and added a small fix
1. check for windows OS
2. print out stderr as well

hoping that is ok :)

royhadad (approved):

royhadad:
@amustaque97 Thank you for the contribution, awesome work!! 🚀🚀🚀

PR #813, opened 2022-10-18
feat: add docs command for datree #763 

### LINKED ISSUE #763: add documentation command
An easy way to open datree documentation

Adding `documentation` or `docs` command that will open a browser with datree documentation

imrushi:
Added `docs` or `documentation` command that will open a browser with datree documentation. Also as per [suggestion](https://github.com/datreeio/datree/issues/763#issuecomment-1225454155) in the issue, I have moved the `OpenBroswer` function to `pkg/utils` package so it can be used to both `test` and `docs` command.

image

### DIFF
--- a/cmd/docs/main.go
+++ b/cmd/docs/main.go
@@ -24,7 +24,7 @@ func New(ctx *DocsCommandContext) *cobra.Command {
 		Short:   "Open datree documentation page",
 		Long:    "Open the default browser with datree documentation page",
 		Aliases: []string{"documentation"},
-		Example: (`
+		Example: utils.Example(`
 		# Open documentation with 'docs'
 		datree docs
 
--- a/cmd/docs/main.go
+++ b/cmd/docs/main.go
@@ -1,16 +1,24 @@
 package docs
 
 import (
-	"fmt"
-	"os"
-
 	"github.com/datreeio/datree/pkg/utils"
 	"github.com/spf13/cobra"
 )
 
-const DEFAULT_ERR_EXIT_CODE = 1
+type DocsCommandContext struct {
+	BrowserCtx utils.OpenBrowserContext
+	URL        string
+}
 
-func New() *cobra.Command {
+func (do *DocsCommandContext) OpenURL(url string) error {
+	err := do.BrowserCtx.OpenBrowser(url)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func New(ctx *DocsCommandContext) *cobra.Command {
 	var docsCmd = &cobra.Command{
 		Use:     "docs",
 		Short:   "Open datree documentation page",
@@ -23,13 +31,15 @@ func New() *cobra.Command {
 		# Open documentation with alias 'documentation'
 		datree documentation
 		`),
-		Run: func(cmd *cobra.Command, args []string) {
-			url := "https://hub.datree.io"
-			err := utils.OpenBrowser(url)
-			if err != nil {
-				fmt.Println(err)
-				os.Exit(DEFAULT_ERR_EXIT_CODE)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if ctx.URL == "" {
+				ctx.URL = "https://hub.datree.io"
 			}
+			err := ctx.BrowserCtx.UrlOpener.OpenURL(ctx.URL)
+			if err != nil {
+				return err
+			}
+			return nil
 		},
 	}
 
--- a/cmd/docs/main.go
+++ b/cmd/docs/main.go
@@ -2,16 +2,19 @@ package docs
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/datreeio/datree/pkg/utils"
 	"github.com/spf13/cobra"
 )
 
+const DEFAULT_ERR_EXIT_CODE = 1
+
 func New() *cobra.Command {
 	var docsCmd = &cobra.Command{
 		Use:     "docs",
-		Short:   "Datree documentation",
-		Long:    "It will open a browser with datree documentation",
+		Short:   "Open datree documentation page",
+		Long:    "Open the default browser with datree documentation page",
 		Aliases: []string{"documentation"},
 		Example: (`
 		# Open documentation with 'docs'
@@ -25,6 +28,7 @@ func New() *cobra.Command {
 			err := utils.OpenBrowser(url)
 			if err != nil {
 				fmt.Println(err)
+				os.Exit(DEFAULT_ERR_EXIT_CODE)
 			}
 		},
 	}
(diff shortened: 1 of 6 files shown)


cmd/docs/main.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/datreeio/datree/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+func New() *cobra.Command {
+	var docsCmd = &cobra.Command{
+		Use:     "docs",
+		Short:   "Datree documentation",
+		Long:    "It will open a browser with datree documentation",
```
myishay: I think the sentence might be better like this: "Open the default browser with datree documentation page"

cmd/docs/main.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/datreeio/datree/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+func New() *cobra.Command {
+	var docsCmd = &cobra.Command{
+		Use:     "docs",
+		Short:   "Datree documentation",
```
myishay: I think the sentence might be better like this: "Open datree documentation page"

cmd/docs/main.go
```diff
@@ -0,0 +1,33 @@
...
+		# Open documentation with 'docs'
+		datree docs
+
+		# Open documentation with alias 'documentation'
+		datree documentation
+		`),
+		Run: func(cmd *cobra.Command, args []string) {
+			url := "https://hub.datree.io"
+			err := utils.OpenBrowser(url)
+			if err != nil {
```
myishay, line 39: What happens if the opening of the browser fails? I think the user would want to receive an error exit code rather than only printing the error. @imrushi WDYT?

cmd/docs/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+			desc:     "Test datree docs",
+			expected: []byte(`Opening https://hub.datree.io in your browser.`),
+		},
+	}
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			cmd := New()
+			out := bytes.NewBufferString("Opening https://hub.datree.io in your browser.")
+			cmd.SetOut(out)
+			err := cmd.Execute()
```
myishay, line 53: This test actually opens my default browser while running it locally. Since I don't know what will be the behavior in headless environment like CI/CD for example, I think the `browser.OpenURL(url)` should be mocked somehow and not actually open a browser in the test.

myishay (commented):
@imrushi  Nice PR!

I added a few comments, lmk what you think

cmd/docs/main.go
imrushi: Okay will update.

cmd/docs/main.go
imrushi: Okay will update.

cmd/docs/main.go
imrushi: Yes, you are right It should get an error exit code. I will update it.

cmd/docs/main_test.go
imrushi: Okay, I will look into how I can mock `browser.OpenURL(url)` behaviour in the test.

imrushi: Hi @myishay, I have checked the way, but I can't find a way to mock. I also checked how the [browser](https://github.com/pkg/browser) library implemented test cases, but they did it the same way and it passes the test case in their CI/CD. 
If you have another idea we can mock, please let me know.

cmd/docs/main.go
```diff
@@ -0,0 +1,37 @@
...
+				fmt.Println(err)
+				os.Exit(DEFAULT_ERR_EXIT_CODE)
```
myishay: I think the `os.Exit`  should be managed in one place ([here](https://github.com/datreeio/datree/blob/2c8a8a649b692e87a07f77c03db2b6b1af1f7d35/main.go#L71)), and here we should just return the error as in other places in which we are using `RunE` cobra method

cmd/docs/main_test.go
myishay: Looking at it now

myishay: > Hi @myishay, I have checked the way, but I can't find a way to mock. I also checked how the [browser](https://github.com/pkg/browser) library implemented test cases, but they did it the same way and it passes the test case in their CI/CD.  If you have another idea we can mock, please let me know.

I looked at the browser library. It seems like they don't have any tests:
Screen Shot 2022-10-20 at 14 29 29
By the way, you can have a look at the [tests workflow run](https://github.com/datreeio/datree/actions/runs/3287564623/jobs/5416874766) and see that it failed probably because it is a "headless machine".

Regarding mocking functions, there are 2 options for that which are described pretty good in the top answer of this [stack-overflow question](https://stackoverflow.com/questions/19167970/mock-functions-in-go).
The solution which we used in other commands for mocking is creating context for each command and then mocking this context in the test file. I think the easiest implementation in our code is [in the version command test case](https://github.com/datreeio/datree/blob/main/cmd/version/main_test.go).
It might not be the easiest to implement for the first time, so don't hesitate to ask questions.

imrushi: Thank you for showing me the way. I will try my best.

cmd/docs/main.go
imrushi: Done

cmd/docs/main_test.go
imrushi: Hi @myishay, I have created the mock test could you please review it and let me know if I have done anything wrong all test cases are passed without opening the browser.

cmd/docs/main.go
```diff
@@ -0,0 +1,47 @@
...
+	return nil
+}
+
+func New(ctx *DocsCommandContext) *cobra.Command {
+	var docsCmd = &cobra.Command{
+		Use:     "docs",
+		Short:   "Open datree documentation page",
+		Long:    "Open the default browser with datree documentation page",
+		Aliases: []string{"documentation"},
+		Example: (`
```
TzlilSwimmer123: can you please use `utils.Example` like used in `publish/main.go`?

TzlilSwimmer123 (changes requested):

cmd/docs/main.go
myishay: @TzlilSwimmer123 I missed that one, thanks!
@imrushi I think we can merge after you fix this one 🚀

imrushi: @myishay Please check I have made the change.

TzlilSwimmer123 (approved):

myishay (approved):

PR #818, opened 2022-10-24
feat: add octupos to the list of ci

TzlilSwimmer123:
windows env vars:
Screen Shot 2022-10-24 at 13 38 38

ubuntu env vars:
Screen Shot 2022-10-24 at 13 38 45

### DIFF
--- a/pkg/ciContext/ciContext.go
+++ b/pkg/ciContext/ciContext.go
@@ -43,21 +43,25 @@ var ciMap = map[string]*CIMetadata{
 	"SAIL_CI":                            {CIEnvValue: "sail", ShouldHideEmojis: false},
 }
 
-var ciMapPrefix = map[string]CIMetadata{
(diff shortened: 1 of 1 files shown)


royhadad (approved):

pkg/ciContext/ciContext.go
```diff
@@ -70,22 +74,28 @@ func Extract() *CIContext {
...
 	}
 
-	return ""
+	for env := range ciFallbackEnvVar {
+		if isKeyInEnv(env) {
+			return env, ciFallbackEnvVar[env]
+		}
+	}
+
+	return "", nil
```
myishay, line 98: just making sure, could this possibly lead us to `nil pointer` error in some cases? @TzlilSwimmer123 @royhadad

TzlilSwimmer123: No because who consumes it checks if `isCI := len(ciName) > 0`

PR #822, opened 2022-10-26
feat: cli add save-rendered flag

shalev007:
- feat: add save rendered flag to test command
- gix: move save rendered to executor for testing
- test: add flag test

### DIFF
--- a/cmd/test/main.go
+++ b/cmd/test/main.go
@@ -49,6 +49,11 @@ type K8sValidator interface {
 	GetK8sFiles(filesConfigurationsChan chan *extractor.FileConfigurations, concurrency int) (chan *extractor.FileConfigurations, chan *extractor.FileConfigurations)
 }
 
+const (
+	SAVE_RENDERED_YES string = "YES"
+	SAVE_RENDERED_NO  string = "NO"
+)
+
 type TestCommandFlags struct {
 	Output               string
 	K8sVersion           string
@@ -60,6 +65,7 @@ type TestCommandFlags struct {
 	PolicyConfig         string
 	NoRecord             bool
 	SkipValidation       string
+	SaveRendered         string
 }
 
 // TestCommandFlags constructor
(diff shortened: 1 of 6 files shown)


cmd/test/main.go
```diff
@@ -261,6 +274,7 @@ func (flags *TestCommandFlags) AddFlags(cmd *cobra.Command) {
 	// kubeconform flag
 	cmd.Flags().StringArrayVarP(&flags.SchemaLocations, "schema-location", "", []string{}, "Override schemas location search path (can be specified multiple times)")
 	cmd.Flags().BoolVarP(&flags.IgnoreMissingSchemas, "ignore-missing-schemas", "", false, "Ignore missing schemas when executing schema validation step")
+	cmd.Flags().StringVarP(&flags.SaveRendered, "save-rendered", "", "", "Save The rendered file. needs to be a path to a directory")
```
myishay: does this have to be a path to a directory? or could it also be a path to a yaml file?

myishay: also, should this be global path only? or also local path?

cmd/test/main.go
```diff
@@ -357,16 +372,30 @@ func TestWrapper(ctx *TestCommandContext, args []string, testCommandFlags *TestC
...
+		if testCommandData.SaveRendered != "" {
+			content := []byte{}
+			_, err := os.Stdin.Read(content)
+			if err != nil {
+				return err
+			}
 
-		if _, err := io.Copy(tempFile, os.Stdin); err != nil {
-			return err
+			savedFilenam, err := ctx.CommandRunner.SaveRenderedFile(testCommandData.SaveRendered, content)
```
myishay: I think this `SaveRenderedFile` function can receive `io.Reader` as an argument type instead of the content `[]byte`.
This way you don't have to save anything in memory. Think of huge helm charts that have to be saved in memory just to be written again in the provided path.

myishay (commented):
I have 1 comment :)

Another issue:
what happens if I use `datree test <filename> --rendered-file="<another filename>"`?

myishay (approved):

PR #834, opened 2022-11-10
830-fix-datree-offline-mode-error

adamdepollo:
Add "blocked" to the list of network errors to resolve issue with execution in an airgapped environment.

### DIFF
--- a/pkg/utils/errors.go
+++ b/pkg/utils/errors.go
@@ -18,7 +18,7 @@ func ParseErrorToString(err interface{}) string {
 }
 
 func IsNetworkError(err error) bool {
-	networkErrors := []string{"network error", "connection refused", "no such host", "i/o timeout", "server misbehaving"}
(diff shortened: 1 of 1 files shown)


pkg/utils/errors.go
```diff
@@ -18,7 +18,7 @@ func ParseErrorToString(err interface{}) string {
 }
 
 func IsNetworkError(err error) bool {
-	networkErrors := []string{"network error", "connection refused", "no such host", "i/o timeout", "server misbehaving"}
+	networkErrors := []string{"network error", "connection refused", "no such host", "i/o timeout", "server misbehaving", "blocked"}
```
royhadad, line 21: Did you test it out? I think it won't work because of case sensitivity

adamdepollo: sorry this is a little premature :sweat_smile: was testing out the Github CLI and raised this before I meant to, I'll close it and update once I have everything working

PR #864, opened 2023-01-16
fix: require key to prevent false negatives in rule 52

hadar-co:


pkg/defaultRules/defaultRules.yaml
```diff
@@ -1602,6 +1602,8 @@ rules:
               kind:
                 enum:
                   - Pod
+            required:
+              - kind
```
royhadad, line 1606: only here? 😂

royhadad (approved):

PR #871, opened 2023-01-25
fix: panic errors in extractConfigurationK8sData

royhadad:
when the resources "annotations" is an array instead of an object, we get this error:

Screenshot 2023-01-25 at 13 11 34


pkg/extractor/extractor.go
```diff
@@ -131,22 +131,22 @@ func extractConfigurationK8sData(content []byte, yamlNode yaml.Node) Configurati
...
 
-	if apiVersion := jsonObject["apiVersion"]; apiVersion != nil {
-		configuration.ApiVersion = apiVersion.(string)
+	if apiVersion, ok := jsonObject["apiVersion"].(string); ok {
+		configuration.ApiVersion = apiVersion
 	}
 
-	if kind := jsonObject["kind"]; kind != nil {
-		configuration.Kind = kind.(string)
+	if kind, ok := jsonObject["kind"].(string); ok {
```
TzlilSwimmer123, line 148: it means that we don't need the check of `kind != nil`? because of the use of `ok`?

royhadad: yeah it's fine :)

pkg/extractor/extractorTestFiles/annotationsAreArray.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: rss-site
+  annotations:
+    - aaa: bbb
```
shmu3l, line 6: lets just make sure skip annotations is passed.

shmu3l (approved):

PR #876, opened 2023-02-12
feat: add CIS policy and new rules

hadar-co:


pkg/defaultRules/defaultRules.yaml
```diff
@@ -2987,3 +3006,431 @@ rules:
...
                     - CSIStorageCapacity
+  - id: 95
+    name: Prevent use of the `cluster-admin` role
+    uniqueName: CIS_INVALID_ROLE_CLUSTER_ADMIN
+    enabledByDefault: false
+    documentationUrl: "https://hub.datree.io/built-in-rules/prevent-cluster-admin-role"
+    messageOnFailure: "Incorrect value for key `name` - the RBAC role `cluster-admin` provides wide-ranging powers over the environment and should be used only where needed"
+    category: CIS
+    complexity: easy
+    impact: Deploying a resource with a deprecated API version will cause Kubernetes to reject it
```
royhadad: .

pkg/defaultRules/defaultRules.yaml
```diff
@@ -2987,3 +3006,431 @@ rules:
...
+  - id: 96
+    name: Prevent access to secrets
+    uniqueName: CIS_INVALID_VERB_SECRETS
+    enabledByDefault: false
+    documentationUrl: "https://hub.datree.io/built-in-rules/prevent-access-to-secrets"
+    messageOnFailure: "Incorrect value/s for key `verbs` - access to secrets should be restricted to the smallest possible group of users to reduce the risk of privilege escalation"
+    category: CIS
+    complexity: medium
+    impact: Inappropriate access to cluster secrets can allow an attacker to gain access to the cluster or external resources whose credentials are stored as secrets
+    schema:
```
royhadad, line 3044: fix schema and unit tests

pkg/policy/tests/102-pass.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  creationTimestamp: 2019-06-16T00:12:34Z
+  name: build-robot
```
royhadad, line 5: name: default

pkg/policy/tests/103-pass.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Pod
+spec:
+  containers:
+  - name: my-pod
+    image: gcr.io/google-samples/node-hello:1.0
+    securityContext:
+      capabilities:
+        drop: ["NET_RAW"]
```
royhadad:
```suggestion
        drop: ["NET_RAW", "BlaBla"]
```

pkg/policy/tests/102-pass.yaml
hadar-co: decided to leave it as is, I think it's better to ensure that without the name the rule doesn't fail. If we set the name to `default` then we need to add the key with the right value to make it pass

royhadad (approved):

PR #881, opened 2023-02-20
feat: starter as the default policy

shalev007:


pkg/defaultPolicies/defaultPolicies.yaml
```diff
@@ -317,6 +316,7 @@ policies:
       - identifier: ALL_EXPOSED_SECRET_PRIVATEKEY
         messageOnFailure: Secret data found in config - keep your sensitive data elsewhere to prevent it from being stolen
   - name: Starter
+    isDefault: true
```
royhadad, line 319: What about hardcoded places that search for the word "Default" in the CLI/webhook/backend/front?

royhadad (approved):

Issue #642, opened 2022-05-24
datree requires policy-as-code mode to be enabled even in offline mode


raphaelehret:
When running tests in offline mode with a policy file specified, datree still asks to be logged in and have policy-as-code mode enabled, which conflicts with the offline mode use-case

**Steps to reproduce the behavior:**
Activate offline mode, then try to run a test with policy config specified
```
datree config set offline local
datree test file.yaml --policy-config ./policies.yaml
```

**Expected behavior**
Test runs with policies specifed in the `policies.yaml` file, as specified in [the documentation](https://hub.datree.io/setup/offline-mode)

**Actual behavior**
Command fails with the following error: 
```
to use --policy-config flag you must first enable policy-as-code mode: https://hub.datree.io/policy-as-code
```
as if it was still trying to run in online mode

**Desktop (please complete the following information):**
I tried with both 
- Mac
- Debian

**Datree version (run `datree version`):**
I tried with both 
- Version: 1.4.28
 - Version: 1.4.29-rc

**Client ID (`cat ~/.datree/config.yaml`):**
client_id: TeasQHTr6uJQHU8pqPyUUD
offline: local
token:  XXX

eyarz:
I'm not able to reproduce the bug :/
![image](https://user-images.githubusercontent.com/19731161/170065064-e36312e9-b062-4517-b166-eb76e07db8a0.png)

if you enable policy as code mode from your dashboard and run this:
```
datree test file.yaml --policy-config ./policies.yaml
```
will datree perform the schema validation check?
![image](https://user-images.githubusercontent.com/19731161/170065750-1b294101-30af-4d2e-809f-19ae1e363cab.png)

raphaelehret:
> if you enable policy as code mode from your dashboard and run this:
``` 
datree test file.yaml --policy-config ./policies.yaml
```
> will datree perform the schema validation check?

Yes, it does work, and I also figured out another behavior. 
If I try offline mode **with my internet disabled**, datree will use my local policy. However if I try offline mode **with my internet enabled**, datree will ask for polices-as-code to be enabled. 

So as of right now, offline mode isn't actually offline unless you have your internet disabled. Is this desired behavior? From my understanding, offline-mode should be offline regardless of my connectivity status

eyarz:
@raphaelehret this is actually the expected behavior.
may I ask what is the use case that you have that requires you to run datree in offline mode although you have an internet connection?

raphaelehret:
[...]

PR #897, opened 2023-03-08
feat: allow offline mode when connected to the internet

### LINKED ISSUE #890: Cannot use Policy as code from CLI
**Describe the bug**
Running datree from cli only (or helm plugin)
I canno use a local policy file, still getting an error

**To Reproduce**
Steps to reproduce the behavior from any helm chart folder :
1. `helm datree config set offline local`
2. `helm datree test . --verbose --no-record --policy-config datree/policy.yaml`

output is:
```
to use --policy-config flag you must first enable policy-as-code mode: https://hub.datree.io/policy-as-code
```
**Expected behavior**
datree run from local file policy

I'm running datree from gitlab CI so I do not want any connection to the web ui not internet connection.
I have so skip some test since some new addtition are not backward compatible with current produztion setup.
I've downloaded default policy.yaml and wnat to intruct datree to run from it keeping the offline mode on.

zmotu:
This pull request aims to address the use case where

> CI environments that may not be air gapped (ie. have internet access) but do not want the additional dependency on the external API introduced in their CI pipelines either for security, reproducibility or reliability reasons.

This use case was originally discussed in 
- https://github.com/datreeio/datree/issues/642
- https://github.com/datreeio/datree/issues/890

### DIFF
--- a/pkg/cliClient/cliClient_test.go
+++ b/pkg/cliClient/cliClient_test.go
@@ -132,11 +132,10 @@ func TestRequestEvaluationPrerunDataSuccess(t *testing.T) {
 		test_requestEvaluationPrerunData_anonymousSuccess(),
 	}
 
-	httpClientMock := mockHTTPClient{}
-
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			body, _ := json.Marshal(tt.mock.response.body)
+			httpClientMock := mockHTTPClient{}
 			mockedHTTPResponse := httpClient.Response{StatusCode: tt.mock.response.status, Body: body}
 			httpClientMock.On("Request", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(mockedHTTPResponse, tt.mock.response.error).Once()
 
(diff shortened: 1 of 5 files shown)


pkg/cliClient/cliClient_test.go
```diff
@@ -132,11 +132,10 @@ func TestRequestEvaluationPrerunDataSuccess(t *testing.T) {
 		test_requestEvaluationPrerunData_anonymousSuccess(),
 	}
 
-	httpClientMock := mockHTTPClient{}
-
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			body, _ := json.Marshal(tt.mock.response.body)
+			httpClientMock := mockHTTPClient{}
```
zmotu, line 138: Moving this line into the loop addresses a bug in the tests where the call to Mock.AssertCalled was hitting the following edge case outlined in the Mock.AssertCalled comments. 

```
// AssertCalled asserts that the method was called.
// It can produce a false result when an argument is a pointer type and the underlying value changed after calling the mocked method.
func (m *Mock) AssertCalled(t TestingT, methodName string, arguments ...interface{}) bool {

```


The impact of this was that the tests originally expected a cliClient request to be called when "offlineMode=local" but the assertion was not failing. The test case for offline mode has now been broken into a seperate test to allow for a Mock.AssertNotCalled check to be made which is now the expected behaviour.

adifayer:
@zmotu Thank you for submitting this PR! We will look into it and provide our review in the next couple of days🤩

royhadad (approved):
Looking good! 🚀🚀
I added a message to let users know whenever they're running in offline mode

image

eyarz:
docs is also updated: https://github.com/datreeio/docs/pull/153

PR #899, opened 2023-03-09
feat: support rego in custom rules

TzlilSwimmer123:


pkg/jsonSchemaValidator/extensions/customKeyRegoRule.go
```diff
@@ -0,0 +1,123 @@
+// This file defines a custom key to implement the logic for the rule:
```
hadar-co: @TzlilSwimmer123 i think this file should be named "customKeyRegoDefinition" like the others whose name is the same name as the key

go.mod
```diff
@@ -13,46 +13,53 @@ require (
...
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/apimachinery v0.23.5
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require github.com/open-policy-agent/opa v0.49.2
+
 require (
+	github.com/OneOfOne/xxhash v1.2.8 // indirect
```
shmu3l, line 35: are those extra packages are necessary? `xxhash`

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+			return nil
+		} else {
+			return ctx.Error(RegoDefinitionCustomKey, "violation needs to return a boolean")
+		}
+	}
+}
+
+func getPackageFromRegoCode(regoCode string) string {
+	const PACKAGE = "package"
```
shmu3l, line 98: if the word "package" dose not exist in `regoCode`, it can be a panic error?

image

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+)
+
+const RegoDefinitionCustomKey = "regoDefinition"
+
+var regoCodeToEval = RegoDefinition{}
```
shmu3l: im not sure about the global var that all function can modify it

same case but a lil bit different, maybe it can explain better: 
https://gobyexample.com/mutexes

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+	index := strings.Index(regoCode, PACKAGE)
+	// get next single word after "package"
+	packageStr := strings.Fields(regoCode[index:])
+	return packageStr[1]
+}
+
+func retrieveRegoFromSchema(regoCode string) *rego.Rego {
+	mainRegoPackage = getPackageFromRegoCode(regoCode)
+
+	var regoObjectParts []func(r *rego.Rego)
```
shmu3l, line 118: `retrieveRegoFromSchema` - It is assumed that the package is in the libs field of the `regoCode`, but if it is not, would this code still work? It might be better to verify that the `package` key is in the `lib` as well.

same logic as `getPackageFromRegoCode`

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
TzlilSwimmer123: I don't think that panicing here will be good because imagine we have a policy with 6 different rules, 5 are ok and 1 is not, instead of evaluating the rule we will panic the policyCheck, feels like a bad experience

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
TzlilSwimmer123: Basically I need some way to preserve the rego code I get in the Compile function. Basically only 1 place in the code sets and no one updated it. What is the concern?

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
TzlilSwimmer123: Basically the rego code will not compile and in this situation we fail the rule. package `something` is a must

go.mod
royhadad: those seem to be indirect dependencies

TzlilSwimmer123: I run `go mod tidy` and it didn't change

cmd/publish/main.go
```diff
@@ -105,11 +116,46 @@ type MessagesContext struct {
...
+		if !valid {
+			return nil, fmt.Errorf("failed to validate rego custom rules")
+		}
+	}
+
 	return ctx.PublishCliClient.PublishPolicies(policiesConfiguration, localConfigContent.Token)
 }
+
+func validateIfRegoRules(ctx *PublishCommandContext, customRules []interface{}) (bool, error) {
+	for _, regoRule := range customRules {
```
royhadad: at this stage we don't know yet if it's a regoRule
```suggestion
	for _, cusotmRule := range customRules {
```

cmd/publish/regoDefinitionSchema.yaml
```diff
@@ -0,0 +1,19 @@
...
+      regoDefinition:
+        type: object
+        required:
+          - code
+        properties:
+          libs:
+            type: array
+          code:
+            type: string
+            pattern: \bviolation\b
```
royhadad: can we validate that the rule starts with `package xxx` as well?

cmd/publish/regoDefinitionSchema.yaml
```diff
@@ -0,0 +1,19 @@
+definitions:
+  regoPattern:
+    properties:
+      regoDefinition:
+        type: object
+        required:
+          - code
+        properties:
+          libs:
+            type: array
```
royhadad, line 10: validate the item type as well to be a string

cmd/publish/regoDefinitionSchema.yaml
```diff
@@ -0,0 +1,19 @@
+definitions:
```
royhadad, line 1: I was thinking about extending the file `policiesSchema.json` with the rego definition, have you considered it?

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+)
+
+const RegoDefinitionCustomKey = "regoDefinition"
+
+var regoCodeToEval = RegoDefinition{}
+var mainRegoPackage = ""
+var mainModuleFileName = "main.rego"
+var regoFunctionEntryPoint = "violation"
```
royhadad:
```suggestion
const regoFunctionEntryPoint = "violation"
```

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+)
+
+const RegoDefinitionCustomKey = "regoDefinition"
+
+var regoCodeToEval = RegoDefinition{}
+var mainRegoPackage = ""
+var mainModuleFileName = "main.rego"
```
royhadad:
```suggestion
const mainModuleFileName = "main.rego"
```

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+	"strings"
+
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+)
+
+const RegoDefinitionCustomKey = "regoDefinition"
+
+var regoCodeToEval = RegoDefinition{}
+var mainRegoPackage = ""
```
royhadad: This variable is only used in 1 function, it can definitely become a private variable of that function, instead of being exposed to (and potentially mutated by) the entire module

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
royhadad: I share @shmu3l's concern, this code smells funny. 
what's wrong with compiling the rego code in the validation function?
or maybe there is a way to pass a value from the `Compile` function to the `Validate` function?

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+	// Execute the prepared query.
+	rs, err := query.Eval(regoCtx, rego.EvalInput(dataValue))
+
+	if err != nil {
+		return ctx.Error(RegoDefinitionCustomKey, "failed to evaluate rego due to %s", err.Error())
+	}
+
+	if len(rs) != 1 || len(rs[0].Expressions) != 1 {
+		return ctx.Error(RegoDefinitionCustomKey, "failed to evaluate rego, unexpected results")
+	} else {
```
royhadad: a bit easier on the eyes
```suggestion
```

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+		resultsValue := (rs[0].Expressions[0].Value).([]interface{})
+		value, ok := resultsValue[0].(bool)
+		if ok {
+			if value {
```
royhadad: what is this `value`? what does it represent? maybe a different name?

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+func getPackageFromRegoCode(regoCode string) string {
+	const PACKAGE = "package"
+	// find the index of string "package"
+	index := strings.Index(regoCode, PACKAGE)
+	// get next single word after "package"
+	packageStr := strings.Fields(regoCode[index:])
+	return packageStr[1]
```
royhadad: why 1 and not 0?

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return nil, nil
+}
+
+type RegoDefinition struct {
+	Libs []string `json:"libs"`
+	Code string   `json:"code"`
+}
+
+func (s CustomKeyRegoDefinitionSchema) Validate(ctx jsonschema.ValidationContext, dataValue interface{}) error {
```
royhadad: what happens when this `Validate` function throws a panic?
maybe add an error handler as a fallback?
```
defer func() {
		if panicErr := recover(); panicErr != nil {
// do something                 
}
	}()
```

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,116 @@
+// This file defines a custom key to implement the logic for rego rule:
+
+package jsonSchemaValidator
```
royhadad, line 3: ***global comment***
Can we add some test coverage for this feature? (publish, test, --policy-config)
There are a lot of edge cases and I can't really know how the system will behave in each case

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
TzlilSwimmer123: in zero index it's `package`

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
TzlilSwimmer123: so I refactored it, I didn't see we do get it inside `Validate`

cmd/publish/main.go
```diff
@@ -105,11 +116,46 @@ type MessagesContext struct {
...
+
 	return ctx.PublishCliClient.PublishPolicies(policiesConfiguration, localConfigContent.Token)
 }
+
+func validateIfRegoRules(ctx *PublishCommandContext, customRules []interface{}) (bool, error) {
+	for _, customRule := range customRules {
+		b, _ := json.Marshal(customRule)
+		regoRuleStr := string(b)
+
+		isRegoRule := strings.Contains(regoRuleStr, jsonSchemaValidator.RegoDefinitionCustomKey)
```
royhadad: I think this `if` can be removed, to make it simpler. Simply put - we'll validate every rule against the regoDefinitionSchema. if the key is missing - the rule will pass.
then the function can be renamed to `validateRegoDefinitions`

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (customKeyRegoDefinitionSchema CustomKeyRegoDefinitionSchema) Validate(ctx jsonschema.ValidationContext, dataValue interface{}) error {
+	regoDefinitionSchema, err := convertCustomKeyRegoDefinitionSchemaToRegoDefinitionSchema(customKeyRegoDefinitionSchema)
+	if err != nil {
+		return ctx.Error(RegoDefinitionCustomKey, err.Error())
+	}
+
+	regoCtx := context.Background()
+
+	r, err := retrieveRegoFromSchema(regoDefinitionSchema)
```
royhadad: please rename this `r`

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
```diff
@@ -0,0 +1,146 @@
...
+const RegoDefinitionCustomKey = "regoDefinition"
+
+type CustomKeyRegoDefinitionCompiler struct{}
+
+type CustomKeyRegoDefinitionSchema map[string]interface{}
+
+var CustomKeyRegoRule = jsonschema.MustCompileString("customKeyRegoDefinition.json", `{
+	"properties" : {
+		"regoDefinition": {
+			"type": "object"
```
royhadad, line 24: Have you considered putting the `regoDefinitionSchema.yaml` here?

royhadad (dismissed):

pkg/jsonSchemaValidator/extensions/customKeyRegoDefinition.go
TzlilSwimmer123: I think that atm it is connected only to publish command so I feel good to leave it there

royhadad (approved):

hadar-co (approved):

PR #901, opened 2023-03-13
feat: add datree-system pre-commit hook

noorul:
There are scenarios were one would like to use datree that is already installed in the system. Provide an alternate hook datree-system to use the datree binary install in the system.


adifayer:
@noorul Thank you for raising this PR! We'll review it in the upcoming couple of days :)

eyarz (approved):

PR #909, opened 2023-03-28
feat: add category and rules for removed APIs

hadar-co:


pkg/defaultRules/defaultRules.yaml
royhadad, line 1: doesn't this belong to the category "deprecation"?

hadar-co: nope, new category for removed APIs

royhadad (approved):

Issue #737, opened 2022-07-24
Configure a custom "default" schema location and policy config for air gapped environments


uriahrokach:
**Is your feature request related to a problem? Please describe.**

We want to use datree in our air - gapped environment, and so far it's working great, but every time we use it we have to specify `--schema-location` and `--policy-config` and then specify our own configuration. 

We would like to write a configuration of our own, upload it to git, and then configure it as our default  policy, so that we don't have to specify every time. 

This would also benfit other teams in our company that would like to use datree, for which we can configure our own policy and then distribute our configuration.

**Describe the solution you'd like**

There are two good possible solutions in my opinion: The first being two environment variables for the `DATREE_SCHEMA_LOCATION` and `DATREE_POLICY_CONFIG`. The second can be configuring this using `datree config` command.

**Describe alternatives you've considered**

We have have thought of creating an alias or a bash function for the datree command that contains this, but this would be much better "out of the box".

adifayer:
Great idea! 
Here are the product requirements for who ever wants to grab this one:
1. Configure `--policy-config` and `--schema-location` values by adding it to the config file using the `datree set config` command or editing the config yaml file itself.
2. Order for reading values-
CLI flag command > Env variable > Config file

Meyazhagan:
@adifayer I would like to contribute on this enhancement. And I am working on it.

eyarz:
@Meyazhagan you got it.

PR #927, opened 2023-04-11
feat:add configuration for schema_locations, policy_config in config

Meyazhagan:
related issue - #737

### DIFF
--- a/cmd/test/main.go
+++ b/cmd/test/main.go
@@ -298,19 +298,19 @@ func GenerateTestCommandData(testCommandFlags *TestCommandFlags, localConfigCont
 
 	var policies *defaultPolicies.EvaluationPrerunPolicies
 	var err error
-	policyConfigEnv := os.Getenv(DatreePolicyConfig)
 
-	if testCommandFlags.PolicyConfig != "" || localConfigContent.PolicyConfig != "" || policyConfigEnv != "" {
+	var policyConfig string
+	if testCommandFlags.PolicyConfig != "" {
+		policyConfig = testCommandFlags.PolicyConfig
+	} else if policyConfigEnv, ok := os.LookupEnv(DatreePolicyConfig); ok {
+		policyConfig = policyConfigEnv
+	} else if localConfigContent.PolicyConfig != "" {
+		policyConfig = localConfigContent.PolicyConfig
+	}
+
+	if policyConfig != "" {
 		if localConfigContent.Offline != "local" && !evaluationPrerunDataResp.IsPolicyAsCodeMode {
 			return nil, fmt.Errorf("to use custom policy-config you must first enable policy-as-code mode: https://hub.datree.io/policy-as-code")
-		}
-
-		policyConfig := testCommandFlags.PolicyConfig
-		if policyConfig == "" {
-			policyConfig = policyConfigEnv
-		}
-		if policyConfig == "" {
-			policyConfig = localConfigContent.PolicyConfig
 		}
 
 		policies, err = policy.GetPoliciesFileFromPath(policyConfig)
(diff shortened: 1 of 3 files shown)


adifayer:
Thank you for raising this PR! @Meyazhagan We will look into it in the upcoming days 🤩

pkg/localConfig/localConfig.go
```diff
@@ -101,14 +108,29 @@ func (lc *LocalConfigClient) Set(key string, value string) error {
 		return err
 	}
 
-	viper.Set(key, value)
+	tranformedValue := transformValue(key, value)
```
niv1612: The name `transformValue` isn’t indicative. I would suggest simply writing an inline if-else statement, something like this:
```
if key == policyConfigKey {
viper.Set(....)
} else if ....
```

cmd/test/main.go
```diff
@@ -293,13 +298,22 @@ func GenerateTestCommandData(testCommandFlags *TestCommandFlags, localConfigCont
 
 	var policies *defaultPolicies.EvaluationPrerunPolicies
 	var err error
+	policyConfigEnv := os.Getenv(DatreePolicyConfig)
```
niv1612: Can you please explain why the environment variable is being used in this context? I’m not quite sure of its purpose and would appreciate some clarification.

cmd/test/main.go
```diff
@@ -293,13 +298,22 @@ func GenerateTestCommandData(testCommandFlags *TestCommandFlags, localConfigCont
...
 
-	if testCommandFlags.PolicyConfig != "" {
+	if testCommandFlags.PolicyConfig != "" || localConfigContent.PolicyConfig != "" || policyConfigEnv != "" {
 		if localConfigContent.Offline != "local" && !evaluationPrerunDataResp.IsPolicyAsCodeMode {
-			return nil, fmt.Errorf("to use --policy-config flag you must first enable policy-as-code mode: https://hub.datree.io/policy-as-code")
+			return nil, fmt.Errorf("to use custom policy-config you must first enable policy-as-code mode: https://hub.datree.io/policy-as-code")
+		}
+
+		policyConfig := testCommandFlags.PolicyConfig
+		if policyConfig == "" {
```
niv1612: I noticed that there are two identical if statements. Was it made by mistake? It appears to be a potential bug

cmd/test/main.go
```diff
@@ -317,14 +331,23 @@ func GenerateTestCommandData(testCommandFlags *TestCommandFlags, localConfigCont
 		return nil, err
 	}
 
+	schemaLocations := testCommandFlags.SchemaLocations
+	if len(schemaLocations) == 0 {
```
niv1612: Same here (two identical if statements). Was it made by mistake? If not, could you explain?

niv1612 (changes requested):

cmd/test/main.go
Meyazhagan: Not sure why I went with this approach. I was looking for `??` in go lang, wonder go was not support. May be that influenced.

I rewrote these conditions. Take a look at.

pkg/localConfig/localConfig.go
Meyazhagan: I too love this simple way. I changed this to inline if-else

cmd/test/main.go
niv1612: @Meyazhagan  I see that you are still using the environment variable for policyConfig. Can you explain why? We don't use environment variables currently in this context.

Meyazhagan: @niv1612 the user described, the solution to set policyConfig, either by config or env. 

> Order for reading values-
CLI flag command > Env variable > Config file

I didn't find a place for env wrapper like we have for localConfig. So I used it directly here.

niv1612 (approved):

PR #938, opened 2023-05-15
feat: issue#647 export log to file

### LINKED ISSUE #647: Write policy check results to local log file
Background:
1. Github provides an option to [add a job summary within workflows](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) for a nice stylized output. We would like to implement this within our github action.
2. Writing policy check results to a log file will allow us to understand and debug issues more easily.

Solution:
Each "datree test" invocation will write its output to a log file (in JSON format) in addition to the output displayed in the CLI.
The log files will reside in an OS temp folder where they will be deleted periodically.

Our github action will read this log file to extract details of the policy check (num of failed rules, num of files scanned, etc.) and use them in the job summary.

alonzyl:
I would appreciate your feedback on whether there are any areas that I could enhance this cr. Any suggestions for improvement would be greatly appreciated.

while writing the code i came across a few questions about the flow, I would be happy to change anything if you think it would be better. 
- on validation should i try to write in the directory to check if there are righting rights?
- should all the test fail if the directory is not valid?
- maybe not limit the file saving only to json, allow all supported output types.


hadar-co:
Hey @alonzyl :)

**on validation should i try to write in the directory to check if there are righting rights?
should all the test fail if the directory is not valid?**
Regarding these 2 questions, you should check if the directory exists and is writable. See [here](https://stackoverflow.com/questions/20026320/how-to-tell-if-folder-exists-and-is-writable) for some approaches to achieve this. If the dir does not exist or is not writable the export should be skipped without any other side-effects.

**maybe not limit the file saving only to json, allow all supported output types.**
I think JSON is enough, it is flexible enough to cover most use-cases, if we see a need for more output types we can add them in the future:)

Thanks!

alonzyl:
Hi, I updated the code with the comments above. Would like to hear you opinion :)

adifayer:
@alonzyl Thank you for submitting this PR! 🚀 We will review it in the upcoming days :)

cmd/test/main.go
```diff
@@ -472,7 +477,25 @@ func test(ctx *TestCommandContext, paths []string, testCommandData *TestCommandD
 		CliVersion:            ctx.CliVersion,
 		IsCI:                  ctx.CiContext.IsCI,
 		Quiet:                 testCommandData.Quiet,
-	})
+	}
+	err = evaluation.PrintResults(evaluationData)
+
+	if testCommandData.SaveResults != "" {
+		if evaluation.IsWritableDirectory(testCommandData.SaveResults) {
```
shalev007: This function looks a little redundant to me, I mean why not just try to `writeFile` and if an error is returned then just return the error?

The error returned from this should look like this, which is understandable enough, WDYT?
@alonzyl 
image

cmd/test/main.go
```diff
@@ -472,7 +477,25 @@ func test(ctx *TestCommandContext, paths []string, testCommandData *TestCommandD
...
+			resultsText, err := evaluation.GetjsonResult(evaluationData)
+			if err != nil {
+				return err
+			}
+
+			err = ioutil.WriteFile(testCommandData.SaveResults, []byte(resultsText), 0666)
+			if err != nil {
+				fmt.Println(err)
```
shalev007: This print seems a bit redundant as well, since the error returned will be printed out

pkg/evaluation/save_results_flag.go
```diff
@@ -0,0 +1,36 @@
+package evaluation
+
+import (
+	"os"
+	"path/filepath"
+
+	"golang.org/x/sys/unix"
+)
+
+func IsWritableDirectory(filePath string) bool {
```
shalev007: If you indeed don't use this function, This whole file can be deleted as well, less code makes it simpler

cmd/test/main_test.go
```diff
@@ -918,6 +919,12 @@ func test_testCommand_no_record_flag(t *testing.T, ctx *TestCommandContext) {
 	assert.Equal(t, ViolationsFoundError, err)
 }
 
+func test_testCommand_save_results_flag(t *testing.T, ctx *TestCommandContext) {
+	flags := TestCommandFlags{SaveResults: "./"}
+	err := flags.Validate()
+	assert.NoError(t, err)
```
shalev007, line 925: **Optional** - You could add a test case for when there is a wrong path and an error is thrown

shalev007:
Hey @alonzyl thanks for the contribution! 🙏 
looks simple and good however I think there's some extra code there,  take a look at the comments

alonzyl:
Hi, I totally agree with your remarks and have made the necessary changes accordingly. If there are more things that I can improve, I would be happy to make further changes :)

shalev007 (approved):

shalev007:
@alonzyl Thanks you so much for the changes, it looks great.
merged 🥳

PR #944, opened 2023-06-07
perf: cache compiled rules to improve performance by 30%

royhadad:
Note: this is a quick win. If we need to improve performance even further in the future, we should use proper performance profiling, and implement some bigger refactors :)

Scanner performance with ~200 resources and 62 rules enabled
Before:
Screenshot 2023-06-07 at 10 44 19

After:
Screenshot 2023-06-07 at 10 46 32

### DIFF
--- a/pkg/jsonSchemaValidator/validator.go
+++ b/pkg/jsonSchemaValidator/validator.go
@@ -3,14 +3,15 @@ package jsonSchemaValidator
 import (
 	"encoding/json"
 	"fmt"
-	extensions "github.com/datreeio/datree/pkg/jsonSchemaValidator/extensions"
-	"github.com/ghodss/yaml"
-	"github.com/santhosh-tekuri/jsonschema/v5"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
+
+	extensions "github.com/datreeio/datree/pkg/jsonSchemaValidator/extensions"
+	"github.com/ghodss/yaml"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 type JSONSchemaValidator struct {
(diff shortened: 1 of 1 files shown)


pkg/jsonSchemaValidator/validator.go
```diff
@@ -70,12 +74,21 @@ func (jsv *JSONSchemaValidator) Validate(schemaContent string, yamlContent []byt
 	compiler.RegisterExtension("customKeyRule101", extensions.CustomKeyRule101, extensions.CustomKeyRule101Compiler{})
 	compiler.RegisterExtension("customKeyRegoRule", extensions.CustomKeyRegoRule, extensions.CustomKeyRegoDefinitionCompiler{})
 
-	schema, err := compiler.Compile("schema.json")
-	if err != nil {
-		return nil, err
+	// compiler.Compile() is an expensive operation. We cache the compiled schema in rulesSchemasCache to avoid re-compiling the same schema.
+	if _, ok := jsv.rulesSchemasCache.Load(schemaContent); !ok {
```
myishay: I think it is better calling the `ok` variable something like `schemaContentExists` or anything similar

pkg/jsonSchemaValidator/validator.go
```diff
@@ -70,12 +74,21 @@ func (jsv *JSONSchemaValidator) Validate(schemaContent string, yamlContent []byt
...
+		schema, err := compiler.Compile("schema.json")
+		if err != nil {
+			return nil, err
+		}
+		jsv.rulesSchemasCache.Store(schemaContent, schema)
+	}
+	schemaAny, ok := jsv.rulesSchemasCache.Load(schemaContent)
```
TzlilSwimmer123: maybe we can use `jsv.rulesSchemasCache.Load(schemaContent)` just once?

myishay (dismissed):

TzlilSwimmer123 (approved):
Looks amazing

pkg/jsonSchemaValidator/validator.go
royhadad: meaningful name vs standard name
both options have benefits and are legit, @TzlilSwimmer123 and I prefer the standard name in this case (`ok`)

PR #954, opened 2023-07-09
fix: exclude knative crd from rule constraints

hadar-co:


royhadad (approved):
loving the unit test! 🙌

PR #955, opened 2023-07-09
fix: remove defaultRules.yaml before running make test

hadar-co:


Makefile
```diff
@@ -20,6 +20,9 @@ build-windows-amd:
 	GOOS=windows GOARCH=amd64 go build -tags $(or $(datree_build_env),staging) -ldflags="-X github.com/datreeio/datree/cmd.CliVersion=1.0.0"
 
 test:
+	# when changing rule logic and/or tests, we want to ensure our new logic is used when running "make test"
```
royhadad:
```suggestion
	# when changing rule logic in the project's defaultRules.yaml, we want to ensure our new logic is used when running "make test"
```

Makefile
```diff
@@ -20,6 +20,9 @@ build-windows-amd:
...
+	# if we don't delete `defaultRules.yaml`, the tests will use the old logic
+	rm -f ~/.datree/defaultRules.yaml
```
royhadad, line 25: does it work without `-f`? probably better if possible

royhadad (dismissed):

Makefile
hadar-co: `-f` is needed - if the file is already deleted, the `rm` command will fail and exit unless `-f` is used

royhadad (approved):

PR #957, opened 2023-07-11
fix: update name rule logic to ignore incorrect kinds

hadar-co:


royhadad (dismissed):

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1891,6 +1891,8 @@ rules:
             not:
               enum:
                 - Kustomization
+                - ClusterRole
```
royhadad: This means `ClusterRole` can omit the `name` field altogether, I think this should be separated into 2 cases:
1. enforce the `name` field for all relevant resources
2. enforce the `name` field format only for some of those resources

royhadad (changes requested):

myishay (dismissed):
👑

pkg/defaultRules/defaultRules.yaml
hadar-co: good catch! @royhadad

royhadad (changes requested):
I see some problems here, call me

pkg/defaultRules/defaultRules.yaml
```diff
@@ -1861,28 +1861,48 @@ rules:
...
+          if:
+            properties:
+              kind:
+                not:
+                  enum:
+                    - ClusterRole
+                    - ClusterRoleBinding
+          then:
+            properties:
+              metadata:
```
royhadad, line 1874: should `metadata` be required?

royhadad (dismissed):

royhadad (approved):

PR #961, opened 2023-07-18
feat: add custom key for cel expression support

TzlilSwimmer123:
Screen Shot 2023-07-19 at 15 09 31


pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
```diff
@@ -0,0 +1,140 @@
...
+	return nil, nil
+}
+
+func (customKeyCELDefinitionSchema CustomKeyCELDefinitionSchema) Validate(ctx jsonschema.ValidationContext, dataValue interface{}) error {
+	CELDefinitionSchema, err := convertCustomKeyCELDefinitionSchemaToCELDefinitionSchema(customKeyCELDefinitionSchema)
+	if err != nil {
+		return ctx.Error(CustomKeyValidationErrorKeyPath, err.Error())
+	}
+	// wrap dataValue (the resource that should be validated) inside a struct with parent object key
+	dataValueObjectWrapper := make(map[string]interface{})
```
shmu3l: naming maybe `resourceWithParentKey` ?

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
```diff
@@ -0,0 +1,140 @@
...
+			return nil, fmt.Errorf("CELExpression must be an object of type CELExpression %s", err.Error())
+		}
+		CELDefinition.CELExpressions = append(CELDefinition.CELExpressions, CELExpression)
+	}
+
+	return &CELDefinition, nil
+}
+
+func getCELEnvInputs(dataValue map[string]interface{}) ([]cel.EnvOption, error) {
+	var inputs map[string]any
```
shmu3l: `var inputs map[string]interface{}`

shmu3l (dismissed):

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
TzlilSwimmer123: I think that it suppose to be any because it can be any type

royhadad: `any` is an alias for `interface{}` :)

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
```diff
@@ -0,0 +1,140 @@
+// This file defines a custom key to implement the logic for rego rule:
```
royhadad:
```suggestion
// This file defines a custom key to implement the logic for a CEL rule:
```

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
TzlilSwimmer123: Good to know:)

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
```diff
@@ -0,0 +1,140 @@
...
+			"type": "array"
+		}
+	}
+}`)
+
+func (CustomKeyCELDefinitionCompiler) Compile(ctx jsonschema.CompilerContext, m map[string]interface{}) (jsonschema.ExtSchema, error) {
+	if customKeyCELRule, ok := m[CELDefinitionCustomKey]; ok {
+		customKeyCELRuleObj, validObject := customKeyCELRule.([]interface{})
+		if !validObject {
+			return nil, fmt.Errorf("CELDefinition must be an object")
```
royhadad:
```suggestion
			return nil, fmt.Errorf("CELDefinition must be an array")
```

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
```diff
@@ -0,0 +1,140 @@
...
+		if res1.Type().TypeName() != "bool" {
+			return ctx.Error(CustomKeyValidationErrorKeyPath, "cel expression needs to return a boolean")
+		}
+
+		celReturnValue, ok := res1.Value().(bool)
+		if !ok {
+			return ctx.Error(CustomKeyValidationErrorKeyPath, "cel expression needs to return a boolean")
+		}
+		if !celReturnValue {
+			return ctx.Error(CustomKeyValidationErrorKeyPath, "cel expression failure message: %s", celExpression.Message)
```
royhadad: "cel expression failure message:"
does @adifayer want this prefix for every CEL failure? how do users experience it?
maybe include a screenshot

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
```diff
@@ -0,0 +1,140 @@
...
+
+type CELDefinition struct {
+	CELExpressions []CELExpression
+}
+
+func convertCustomKeyCELDefinitionSchemaToCELDefinitionSchema(CELDefinitionSchema CustomKeyCELDefinitionSchema) (*CELDefinition, error) {
+	var CELDefinition CELDefinition
+	for _, CELExpressionFromSchema := range CELDefinitionSchema {
+		var CELExpression CELExpression
+		b, err := json.Marshal(CELExpressionFromSchema)
```
royhadad, line 112: why `Marshal` and then `Unmarshal`? if you just need a type conversion then use Go's type conversion. Am I missing something?

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
```diff
@@ -0,0 +1,140 @@
...
+		CELDefinition.CELExpressions = append(CELDefinition.CELExpressions, CELExpression)
+	}
+
+	return &CELDefinition, nil
+}
+
+func getCELEnvInputs(dataValue map[string]interface{}) ([]cel.EnvOption, error) {
+	var inputs map[string]any
+	dataValueBytes, _ := json.Marshal(dataValue)
+	if err := yaml.Unmarshal(dataValueBytes, &inputs); err != nil {
```
royhadad: Let's see if I understand this properly.
- CEL expects the input to be a yaml
- you get it as a Go object
- therefore, you need to convert the object to a yaml
is that correct?

royhadad (dismissed):

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
TzlilSwimmer123: Unfortunately Go's type conversion is not helpful here.  I will try just to make sure I am not mistaking

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
TzlilSwimmer123: Your comment made me understand it was redundant! thanks

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
TzlilSwimmer123: Ok let's see, what is your suggestion?

pkg/jsonSchemaValidator/extensions/customKeyCELDefinition.go
TzlilSwimmer123: I will show it to Adi

hadar-co (approved):

